PR #9, opened 2016-03-03
Added CircleCI testing

mitar:
It does not really work, because running (what is more or less what test runner does internally):

```
export PACKAGE_DIRS=packages
meteor test-packages --once --driver-package 'test-in-console' -p 4096 packages/blaze
```

throws the following error:

```
=> Build failed:                              
   
   While selecting package versions:
   error: No version of blaze satisfies all constraints: @=2.1.4-beta.11, @=2.1.3
   Constraints on package "blaze":
   * blaze@=2.1.4-beta.11 <- top level
   * blaze@=2.1.3 <- top level
   * blaze@2.1.3 <- ui 1.0.8 <- boilerplate-generator 1.0.4 <- webapp 1.2.3 <- autoupdate 1.2.4
   * blaze@2.1.3 <- test-helpers 1.0.5 <- local-test:blaze 2.1.4-beta.11
   
   No version of blaze-tools satisfies all constraints: @=1.0.5-beta.11, @=1.0.4
   Constraints on package "blaze-tools":
   * blaze-tools@=1.0.5-beta.11 <- top level
   * blaze-tools@=1.0.4 <- top level
   
   No version of html-tools satisfies all constraints: @=1.0.6-beta.11, @=1.0.5
   Constraints on package "html-tools":
   * html-tools@=1.0.6-beta.11 <- top level
   * html-tools@=1.0.5 <- top level
   
   No version of htmljs satisfies all constraints: @=1.0.6-beta.11, @=1.0.5
   Constraints on package "htmljs":
   * htmljs@=1.0.6-beta.11 <- top level
   * htmljs@=1.0.5 <- top level
   * htmljs@1.0.5 <- boilerplate-generator 1.0.4 <- webapp 1.2.3 <- autoupdate 1.2.4
   
   No version of spacebars-compiler satisfies all constraints: @=1.0.8-beta.11, @=1.0.7
   Constraints on package "spacebars-compiler":
   * spacebars-compiler@=1.0.8-beta.11 <- top level
   * spacebars-compiler@=1.0.7 <- top level
   * spacebars-compiler@1.0.7 <- boilerplate-generator 1.0.4 <- webapp 1.2.3 <- autoupdate 1.2.4
   
   No version of spacebars satisfies all constraints: @=1.0.8-beta.11, @=1.0.7
   Constraints on package "spacebars":
   * spacebars@=1.0.8-beta.11 <- top level
   * spacebars@=1.0.7 <- top level
   * spacebars@1.0.7 <- boilerplate-generator 1.0.4 <- webapp 1.2.3 <- autoupdate 1.2.4
   
   No version of templating satisfies all constraints: @=1.1.6-beta.11, @=1.1.5
   Constraints on package "templating":
   * templating@=1.1.6-beta.11 <- top level
   * templating@=1.1.5 <- top level
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "tests/test-runner"]
+	path = tests/test-runner
+	url = https://github.com/tozd/meteor-test-runner.git
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,16 @@
+checkout:
+  post:
+    - git submodule sync
+    - git submodule update --init --recursive
+machine:
+  node:
+    version: 0.10.33
+  pre:
+    - curl https://install.meteor.com | /bin/sh
+dependencies:
+  override:
+    - npm install selenium-webdriver
+    - meteor list
+test:
+  override:
+    - tests/test-runner/test-all.sh
--- a/circle.yml
+++ b/circle.yml
@@ -7,10 +7,11 @@ machine:
     version: 0.10.33
   pre:
     - curl https://install.meteor.com | /bin/sh
+    - git clone -b devel https://github.com/meteor/meteor.git /tmp/meteor
 dependencies:
   override:
     - npm install selenium-webdriver
     - meteor list
 test:
   override:
-    - tests/test-runner/test-all.sh
+    - METEOR_WAREHOUSE_DIR=/tmp/meteor tests/test-runner/test-all.sh


mitar:
@martijnwalraven, any suggestion how to fix testing as well?

stubailo:
What if you set PACKAGE_DIRS?

mitar:
I am doing that. See example command above:

```
export PACKAGE_DIRS=packages
meteor test-packages --once --driver-package 'test-in-console' -p 4096 packages/blaze
```

Or do you have something else in mind?

martijnwalraven:
The issue is that you're running `meteor` with the current release version, which is 1.2.1. And because Meteor releases pin the versions of core packages, you get conflicting constraints.

One short term workaround is to run `meteor` with `--release 1.3-beta.11` in the testing script. That is the only version that will work without changing the versions of the split-off Blaze packages though, because they were split off at `beta.11` and version constraints for core packages require an exact match.

This is obviously not a desirable situation of course. I'm looking into loosening the constraints for the split-off packages so they only specify minimal versions. I'm now thinking we should probably try to get this into 1.3, so split-off packages can be on their own release cycle and people can update at will.

I'm encountering some difficulties with removing the split off packages from `meteor/meteor`, because it turns out tools code indirectly depends on Blaze, and building the tool requires local packages.

Another issue I found is that `PACKAGE_DIRS` is currently ignored when running `meteor` from a checkout. So running from checkout is not a solution for testing either. (I also think it is not desirable, and we should probably find a way to get rid of the distinction between release and checkout altogether, so we can be more flexible about what packages we use from source.)

mitar:
I am glad that you are looking into this. Thanks.

stubailo:
ding ding - found a winner:

```
PACKAGE_DIRS=packages curmeteor test-packages packages/*
```

It works great if you run the checkout of Meteor (here, `curmeteor` is an alias to my local Meteor checkout).

So this isn't _ideal_ (especially on CircleCI where you would have to rebuild all of Meteor every time) but it definitely works.

stubailo:
Breaking news: this approach works even without the `PACKAGE_DIRS` variable, because the directories passed to `test-packages` override the core packages.

martijnwalraven:
I tried this as well, and it seems like the best option, but there is a remaining issue with removing the Blaze packages from the `meteor/meteor` repo. We currently have an indirect dependency from the tools code on `blaze`, and because Isopackets do not take `PACKAGE_DIRS` into account, this won't work as is.

mitar:
The downside of this approach is that you test all packages at the same time. This based on my experience can lead to false failures of tests if they interact with each other. This is why I prefer testing each package separately. It would be great if you could specify that Meteor should test only one package, but use other packages for testing as dependencies.

Does Meteor allow you to specify custom testing app?

Does the above approach work also if you run with Meteor 1.2.1? So `meteor test-packages packages/*`?

martijnwalraven:
>  It would be great if you could specify that Meteor should test only one package, but use other packages for testing as dependencies.

Setting `PACKAGE_DIRS` would do just that, so I think this should work:

```
PACKAGE_DIRS=packages curmeteor test-packages packages/blaze
```

mitar:
> PACKAGE_DIRS=packages curmeteor test-packages packages/blaze

That works only when `curmeteor` is at `1.3-beta.11` version, no? Because this is exactly the command I started with:

```
export PACKAGE_DIRS=packages
meteor test-packages --once --driver-package 'test-in-console' -p 4096 packages/blaze
```

martijnwalraven:
@mitar: No, the difference that it assumes you're running `meteor` from a checkout. And running from a checkout does not add the same constraints a regular release uses.

So if you checkout `meteor/meteor` in `~/Development/git/meteor` you could define:
```
alias curmeteor=~/Development/git/meteor/meteor
```

mitar:
So, for tests. Are you suggesting that I checkout meteor and go to `master` branch and use that for testing packages in this repository?

martijnwalraven:
I think that is probably the best we can do for now. Although you may want to use `devel`, at least until 1.3 is released and we merge it to `master`.

mitar:
Are there any changes in testing from `master`?

martijnwalraven:
I'm not sure if the new testing support has also affected `test-packages`. But there might be other dependencies that have changed from `master`, so it seems safest to test against the state of the repository that is closest to when the Blaze packages were split off.

mitar:
But Blaze packages should not depend anymore on anything from core Meteor, no? :-)

martijnwalraven:
Well, I agree we should keep dependencies to a minimum and eliminate them where possible. But there are at still dependencies on `tracker` and related packages that seem hard to get rid off (although we may want to split off those themselves). And as long as the Blaze packages are Meteor packages they will depend on the build tool.

mitar:
Is there I away that I tell normal `meteor` command to do the checkout for me?

I am not sure if `alias meteor...` works with CircleCI.

mitar:
Can I just set `METEOR_WAREHOUSE_DIR` to the path where I cloned Meteor?

mitar:
OK, I think it should work now. Can you enable CircleCI for this repository?

martijnwalraven:
Great, let's give this a try! Should we merge `testing` into `master` first?

mitar:
No, because probably pull requests should also be already tested once you enabled it. We can just close and open this pull request and it should start running the tests.

martijnwalraven:
I think we need `circle.yml` in the main branch in order to run the tests.

mitar:
Have you enabled CircleCI?

martijnwalraven:
Yes, I think so. But I haven't used it before so I'm not sure if I configured it correctly.

mitar:
BTW, is there a reason why you use CircleCI? I do prefer TravisCI for open source projects. CircleCI is nice for private repositories.

martijnwalraven:
I wasn't involved in the decision, so I couldn't tell you the reasons. But it seems like a great service. What is missing for you?

mitar:
I do not know, maybe I am just more used to their structure of the project and tests run pages: https://travis-ci.org/peerlibrary/meteor-blaze-components

In comparison with: https://circleci.com/gh/meteor/blaze

martijnwalraven:
It seems the tests are hanging trying to spawn the test process.

Did you verify the script is working locally first?

mitar:
It seems the error is:

```
Error: Can't get default release version for track undefined
    at [object Object]._.extend.getDefaultReleaseVersion (/tools/packaging/catalog/catalog-remote.js:828:13)
    at Object.release.latestKnown (/tools/packaging/release.js:201:41)
    at /tools/cli/main.js:873:29
```

mitar:
So I try:

```
export PACKAGE_DIRS=packages
export METEOR_WAREHOUSE_DIR=/my/path/to/meteor/clone
meteor test-packages --once --driver-package 'test-in-console' -p 4096 packages/blaze
```

martijnwalraven:
I'm not sure `METEOR_WAREHOUSE_DIR` actually works to run from a checkout (I've never used it). You may need to run `meteor` from the checkout explicitly.

stubailo:
@mitar where is `meteor` coming from? Is that the `meteor` script from the cloned Meteor repo? You shouldn't run the installed Meteor at all in this case.

Also, I think we used CircleCI because of some issues with parallel runs on Travis. We investigated a lot of services and they all had their drawbacks. But maybe for a simpler repo like Blaze going back to Travis will be OK. If it's easier to set up and use I'm all for it.

mitar:
It seems we are moving towards working: https://circleci.com/gh/meteor/blaze/6

Now tests run. There are some issues I cannot reproduce locally. Will have to look into where this is coming from.

mitar:
Done. Testing also supports parallelization, if there will be more containers allocated to the repository.

PR #16, opened 2016-03-31
Template compilation on the server (discussion)

gadicc:
Is there any reason we don't already do this?

I made the changes in this PR and surprisingly everything just worked.  Of course it needs a little more thought.  `Template.body.renderToDocument()` is never called on the server but obviously that code shouldn't be bundled for the server either.

Obviously this is a great help for SSR, e-mail generation, etc.

Actually if I comment out https://github.com/gadicc/meteor-blaze-react-component/blob/master/test-app/server/main.spec.js#L17-L30 and replace it with `import './body.html'`, we get SSR of Blaze templates via React (sorry :)).

I imagine flowrouter-ssr could also add this since they explicitly state that the only reason they don't do it is because Blaze templates aren't available on the server.


mitar:
Yes. I have made similar changes in [Blaze Components](https://github.com/peerlibrary/meteor-blaze-components/issues/110) to get them to work on the server side.

There are some issues. Like that if you make `<body>` or `<head>` in a template on the server, things break: https://github.com/meteor/meteor/issues/5913

So it is not really possible to create isomorphic templates by default. I had to do some more changes: https://github.com/peerlibrary/meteor-blaze-components/commit/6fd24541ca8f6f657e3fc66fb429580f28f0bbca

Ideally, `<head>` and `<body>` for client target would be ignored (or be added dynamically through JavaScript when it loads), while `<head>` and `<body`> on the server target would render it statically to all clients.

gadicc:
Hey, thanks for the quick reply and all the refs.  This all looks awesome, so glad you're on this!  I'm going to amend my blaze-react-component to work with blaze-components over the weekend and make that the official recommendation (I'll mention the HEAD on server issue, yeah, as per your issue on Meteor, the way it's down is quite weird).

mitar:
I like this idea, but I moved it to "future ideas" on the roadmap because I believe first we need to get Blaze itself into a shape for a stand-alone project. But yes, that would be great to have.

I do see here two steps:
* getting server-side to render to string
* getting Meteor to server server-side rendered (Blaze) content on first-load

gadicc:
Yes absolutely, I agree 100%.

I didn't also mention, how happy I am to see Blaze (and related) under your curatorship now... only wish it had happened earlier.

mitar:
Let's start maybe working on this?

So I think we also need tests to make sure everything works.

gadicc:
@mitar, regrettably - from my side at least - I don't foresee any opportunity to work on this for the remainder of this year.  I'm also not using Blaze in any current projects so don't have too much personal incentive in the future either :(  - other than a big past love for Blaze!

I hope you and others will be able to move this issue forward.  I'm delighted that Blaze is functioning in it's own repo now, in community hands.

mitar:
We should probably change this as well: https://github.com/meteor/blaze/blob/master/packages/blaze/attrs.js#L251

mitar:

mitar:
I am closing this pull request for now. Let's have discussion in #26 and then make a new pull request.

mitar:
I am closing this pull request in favor of #223 which describes what should be done. So that maybe somebody else can take this on.

Issue #30, opened 2016-04-28
Repeated class names not rendered correctly


StanLindsey:
https://github.com/meteor/meteor/issues/5632

Original Post
-----
Using Semantic-UI sometimes you need to add the same class name twice, i.e.

<div class="ui two column computer one column mobile grid container">

but this is translated into

`<div class="ui two column computer one mobile grid container">`

breaking the interface.
This is because of DiffingAttributeHandler call this.parseValuewhich reduces the repeated element into one since the element is the key in the array where values are stored

`parseValue: function (attrString) {
    var tokens = {};

    _.each(attrString.split(' '), function(token) {
      if (token)
        tokens[token] = token;
    });
    return tokens;
  }`

Here is the repo to reproduce the error with step by step guide: https://github.com/bitIO/repeated-class-issue

---------
The outcome was of a pull request that appears to fix the issue but lacked tests, comments and was not readable. 
https://github.com/meteor/meteor/pull/5753

mitar:
Can you explain why repeating classes is important?

mitar:
Do you have an idea how this could be optional behavior?

StanLindsey:
I'm just moving issues over as per https://github.com/meteor/blaze/issues/25. Not by priority but from top to bottom to slowly move them over. Maybe we could get some priority labels as well as this one in particular isn't a high priority.

The Semantic-UI Framework sometimes needs repeated classes, which Meteor merges to one instance in the HTML which breaks the UI. Though this is an edge case of CSS use, there may be users who rely on this or other frameworks that adopt this pattern.

Does it need to be optional? Is there a case where duplicate classes are bad and need to be stripped?

mitar:
> Does it need to be optional? Is there a case where duplicate classes are bad and need to be stripped? 

I would say esthetics. :-)

mitar:
And thanks for copying the issue over. I labeled it as low priority.

shadowRR:
@StanLindsey Is this issue supposed to break every time when using the same class twice on the same html tag ? I'm asking because i've just used the `ui middle aligned center aligned grid` on a new app, which use the latest SUI package from atmosphere and Meteor v.1.3.2.4, and it's working, both `aligned` are shown, and SUI work as expected...

StanLindsey:
Well, if you can't replicate it then I guess we have our first closure.

My issue now is we will need to close the meteor/meteor ticket to make things easier for us to move forward with the rest of the issues.

[...]

PR #37, opened 2016-05-04
Handle repeating class names where order matters

talha-asad:
Fix for #30 
Submitted again with a slightly better comment and is no longer a complex one-liner. Has been tested locally and by a few other developers but no actual tests are included.

### DIFF
--- a/packages/blaze/attrs.js
+++ b/packages/blaze/attrs.js
@@ -101,10 +101,18 @@ var ClassHandler = Blaze._DiffingAttributeHandler.extend({
   },
   parseValue: function (attrString) {
     var tokens = {};
-
-    _.each(attrString.split(' '), function(token) {
-      if (token)
-        tokens[token] = token;
+    var similarTokens = 0;
+    
+    _.each(attrString.split(' '), function(token, i) {
+      if (token) {
+        
+        // Incase of duplicate class names where order matters (e.g: Semantic UI), group similar tokens accordingly
+        if (tokens[token] && i > 0) {
+          var groupWithIndex = (i - similarTokens++) - 1;
+          tokens[Object.keys(tokens)[groupWithIndex]] += ' ' + token;
+        } else
+          tokens[token] = token;
+      }
     });
     return tokens;
   }


packages/blaze/attrs.js
```diff
@@ -101,10 +101,18 @@ var ClassHandler = Blaze._DiffingAttributeHandler.extend({
...
-        tokens[token] = token;
+    var similarTokens = 0;
+    
+    _.each(attrString.split(' '), function(token, i) {
+      if (token) {
+        
+        // Incase of duplicate class names where order matters (e.g: Semantic UI), group similar tokens accordingly
+        if (tokens[token] && i > 0) {
+          var groupWithIndex = (i - similarTokens++) - 1;
+          tokens[Object.keys(tokens)[groupWithIndex]] += ' ' + token;
```
mitar: Order of keys in an object is not defined in JavaScript.

mitar:
So, why is this change really required and simply Meteor packages for Semantic UI do not [do the following change](https://github.com/Semantic-Org/Semantic-UI/issues/1526#issuecomment-151857979) can package the package in this way?

talha-asad:
Semantic UI is not at fault here and it makes no sense to have special class syntax for Semantic UI in meteor because blaze needs patching.

Maybe order of keys in an object in javascript is not defined or the same in all cases, but that is a neat hack to fix the issue, which has been working nicely for me for the last few Blaze versions. Perhaps you can rewrite it in a way that will be consistent.

mitar:
> Semantic UI is not at fault here 

From my reading of the HTML5 spec it is:

>  The classes that an HTML element has assigned to it consists of all the classes returned when the value of the class attribute is split on spaces. **(Duplicates are ignored.)**

(Emphasis mine.)

And Meteor is ignoring duplicates per spec.

mitar:
> but that is a neat hack to fix the issue

Yes, sorry, we do not need neat hacks but working code. Readable and no-hacky code with tests is more important than a quick hacky fix.

talha-asad:
After reading the spec, i agree with you. Lets patch Semantic UI ;)

mitar:
So intermediary solution would be that Meteor packages fix this. It seems [people are confirming](https://github.com/Semantic-Org/Semantic-UI/issues/1526#issuecomment-217048865) that a change works. This could be part of the bundling process when packaging semantic UI for Meteor.

Issue #39, opened 2016-05-13
{{elsif}} or {{elseif}} or {{#elseif}}


ramezrafla:
I am suggesting we add an {{elsif}} or {{elseif}} or whatever variation to avoid this ugly piece of code:

```
{{#if condition1}}
{{else}}
  {{#if condition2}}
  {{else}}
     {{#if condition3}}
     {{/if}}
  {{/if}}
{{/if}}
```

Imagine this instead

```
{{#if condition1}}
{{elseif condition2}}
{{elseif condition3}}
{{/if}}
```

I also think this will go a long way in making people more likely to adopt Blaze as it cleans up the code quite a bit for complex apps.

mitar:
Does handlebars have it?

mitar:
The issue is that this might seem a simple addition but would require a lot of changes internally. Like, a lot.

ramezrafla:
Handlebars does not seem to have it.
I was thinking .. a preprocessor would do the trick :)

mitar:
Hmmm. Maybe. So at the level or parser. Like a syntactic sugar. This could work. You could look into spacebars parser if you feel adventurous. :-)

ramezrafla:
Got it :) -- will start digging into it. Is the parser part of this repo or somewhere else?

mitar:
https://github.com/meteor/blaze/tree/master/packages/spacebars-compiler

AdrienTorris:
totally agreed with this issue. Simple if/else if/else is just so usefull

nathantreid:
I've been wanting this everytime I write an ugly if else if else chain, and today it caught my interest, so I've implemented a version using {{elseif}} (example taken from original post):
``` handlebars
{{#if condition1}}
{{elseif condition2}}
{{elseif condition3}}
{{/if}}
```
As suggested, it is simply sugar syntax. The above example is transformed into the compiled spacebars version of this:
``` handlebars
{{#if condition1}}
{{else}}
  {{#if condition2}}
  {{else}}
     {{#if condition3}}
     {{/if}}
  {{/if}}
{{/if}}
```

ramezrafla:
Excellent @nathantreid,
Can you PR with your changes?

nathantreid:
@ramezrafla Check out the PR link above my previous post. :smile:

ramezrafla:
Oops - sorry - I'll blame it on small iPhone screen :) thanks a lot!

capensisma:
[...]

Issue #20, opened 2016-04-17
Get Meteor to use code from this repository for Blaze distributed with Meteor releases


mitar:

laosb:
For now, Blaze doesn't really have a separated versioning. Should we have an independent versioning system?

mitar:
I think it [does](https://github.com/meteor/blaze/blob/master/packages/blaze/package.js#L3) and I also think this was one of the core goals with splitting some (all) packages from Meteor.

laosb:
Every Meteor release gets a bump. That wouldn't be ideal for a **separated** package.

mitar:
I think that would not be needed anymore, this is one of those changes they are working on.

MiroHibler:
I don't think MDG should ship Blaze (or any view layer for that matter) with Meteor in the first place. I'd rather make it all optional as they do now with React since v1.3 for example.

mitar:
I think they will have to do so for foreseeable future (at least they go to Meteor 2.0). And also we will have to keep backwards compatibility for that purpose. We can (as a community) go and do Blaze 2. But we and MDG also have responsibility for current users using Blaze 1.

MiroHibler:
I'm just pessimistic by nature and MDG has been known for abruptly changing their mind ;)

mitar:
OK, let's be more optimistic in the future.

So, let's get community to go through tickets and see which are still relevant. This is only on us. If you are more optimistic about community, then feel free to organize this part.

MiroHibler:
I was (trying to be) sarcastic ;) I wouldn't be here if it wasn't for my optimism! ;)

Lets hear what community has to say.

mitar:
Sorry, a bit long day. Didn't catch that.

brown2rl:
@mitar is there anyone from MDG watching this repo?  I think Blaze 1 has some serious potential.  It would be nice if we could add an option like 

$ meteor upgrade blaze

Essentially the proposed script combines both requirements above, and would pull from the official repo to replace Blaze 1

tcastelli:
+1 Blaze has some good potential inside and I think a symlink to this repo in their own repo would be quite useful to forget shared versioning and dependencies.

mitar:
They are, you can [see them here](https://github.com/meteor/blaze/watchers). :-)

brown2rl:
@mitar ok, cool. let me know what i can do

zol:
[...]

PR #50, opened 2016-07-28
Add {{elseif}} sugar syntax

nathantreid:
Implements #39.

``` handlebars
{{#if counterIsAt 2}}
    test
{{elseif counterIsAt 3}}
    test3
{{elseif counterIsAt 4}}
    test4
{{else}}
    test5
{{/if}}
```

Working example here: https://github.com/nathantreid/blaze-test-issue-39


mitar:
Wohoooo, so great!

kulttuuri:
This would be awesome to have. Great work! 👍

grsabreu:
Any updates on this?

mitar:
Waiting for #20 and that we release first community Blaze version. I would like that one to be without extra features, just bugfixes. And then we can start adding features.

necmettin:
Forgive me if this was discussed (and decided) before, or if we shouldn't discuss PRs; but would it not be nicer, more in-line with Blaze syntax (and easier to read) if we used `{{#elseif` (maybe even `{{#else` at some point?) instead of `{{elseif` ?

mitar:
We should discuss PRs!!! And this is the perfect place.

You are bringing up a great point. But Blaze uses `{{else}}`. So it is a good question which one it should be. I personally vote for `{{elseif}}`.

nathantreid:
For Blaze 1.x, I think it should match the current Blaze if / else structure, so `{{elseif}}` is my vote. 
That said, I actually prefer {{#elseif}} and {{#else}} (I don't think it makes sense to have `{{#elseif}}` with `{{else}}`).
I think it's easier to visually parse the blocks with the `#`.
``` handlebars
{{#if abc}}
  ...
{{#elseif def}}
  ...
{{#else}}
  ...
{{/if}}
```
vs
``` handlebars
{{#if abc}}
  ...
{{elseif def}}
  ...
{{else}}
  ...
{{/if}}
```

mitar:
OK, maybe open an issue with proposal to change `{{else}}` and `{{elseif`}}` to `{{#else}}` and `{{#elseif}}` and I will put it into 2.0 milestone and open it for discussion. Let's see through time how people will upvote or downvote it and see then for 2.0.

exKAZUu:
I think v1 should keep compatibility as much as possible. So I'd like to vote for `{{elseif}}`.

stubailo:
Note that a lot of this syntax is based on handlebars. So it might be interesting to decide whether or not the future of Blaze/Spacebars should include a desire to remain close to that syntax. My hunch would be "no" since probably nobody cares about handlebars these days, and sticking to handlebars was a really big limitation in Blaze for a while.

mitar:
@nathantreid: This is great work. Can you please resolve merge conflicts? And add also all changes to documentation (spacebars API, mention this also somewhere in guide, blaze itself, and so on)?

mitar:
BTW, Handlebars seems to have [else if](https://stackoverflow.com/questions/10736907/handlebars-js-else-if).

See also [example in documentation](http://handlebarsjs.com/block_helpers.html). It seems the syntax is:

```handlebars
{{else if isInactive}}
```

And also:

```handlebars
{{else unless isInactive}}
```

Or in general it seems it can be `{{else <any block helper>}}`.

> It is not necessary to use the same helper in subsequent calls, the unless helper could be used in the else portion as with any other helper. When the helper values are different, the closing mustache should match the opening helper name. 

While I dislike this syntax, I see benefit of having it match Handlebars, and benefit of supporting any block helper. Implementation might be a bit trickier to implement, but this sugar is easy to do:

```handlebars
{{#if foo}}

{{else something bar}}

{{/if}}
```

Becomes:

``
```handlebars
{{#if foo}}

{{else}}
  {{#something bar}}

  {{/something}}
{{/if}}
```

@nathantreid, could you implement it this way?

stubailo:
Yikes.... that seems kind of odd. Are we sure we want to think about handlebars at all? Honestly Handlebars doesn't seem that well designed to start with.

mitar:
For this particular feature I think I would. I would change what `this` means inside a template, but for syntax I would try to stick with Handlebars.

mitar:
Also, this syntax seems to map nicely to the sugar we are trying to make.

nathantreid:
I also think that the {{else if}} syntax looks odd, perhaps we could put that up for a vote?
I personally don't have any desire to match Handlebars going forward, unless of course it's because the syntax is actually makes a lot of sense on its own.

mitar:
It is not just a syntax here, but also a feature. So that you can use any block helper in this way. `elseif` does not support that. I think that a feature of `else <blockname>` is a good one. I use quite some custom block helpers (Blaze Components add extra power to them).

mitar:
@nathantreid, I really love your work here and I think many people would like to see this in. But I also really like that the Handlebars syntax work with arbitrary block helpers. I think that block helpers are really powerful feature of Blaze and with allowing such chaining of them it would be really powerful. Currently block helpers were a bit strange to use because what if you had only two cases, but now you can potentially get multiple cases supported, not just true/false.

Could you please update the pull request to support arbitrary block helpers?

mitar:
So something like this could then work:

```handlebars
{{#tab title="Foo"}}

{{else tab title="Bar"}}

{{else tab title="Baz"}}

{{/tab}}
```

And have a block helper which renders dynamically tabs, opens/closes them, etc.

nathantreid:
I'd be happy too, but I'm swamped right now IRL. If someone else wants to have a go at it, be my guest. Otherwise, I'll try to get to it as soon as I can.

mitar:
No worries. I just wanted to make sure we have a path forward.

mitar:
Any progress here? Anyone else wants to tackle this?

kulttuuri:
This would be fantastic feature to have. @nathantreid still busy with things IRL? :)

mitar:

mitar:
With 2fa3cabcf446a1db571fd63aabea60b974d2c1a0 I implemented the Handlebars syntax for chaining block tags. I am closing this pull request in favor of that.

mitar:
Released as 2.3.0. You have to update templating package.

Issue #51, opened 2016-07-31
Slack join page not working


atom-azimov:
![16 20 51__838x243__heroku nosuchapp-googlechrome](https://cloud.githubusercontent.com/assets/9398439/17276214/8fa51bd0-573b-11e6-845b-753e6fe30004.png)

grsabreu:
Yeah, not working at all :/

MiroHibler:
Hey, @voidale, can you take a look?

alexandesigner:
Somehow the slack spent working on different subdomains... I made a change in the readme or if you would like to be able to ignore and do who have access to the master.

mitar:
Can we find somewhere else to host it? Is this still the simplest way to host a Blaze invite? A Heroku container?

stubailo:
Based on my findings, that's the easiest way - hosting slackin on DO or Heroku

mitar:
And probably for such simple app it should be free to host it on Heroku? So why is it down? @MiroHibler, who has it over? Could more people get access to the domain and the website? I have a bad feeling if only one person has access to all this.

mitar:
Also, who pays for domain?

voidale:
@mitar, @MiroHibler hey guys who of you want to host it? got namecheap account? I will transfer the domain, I will not be taking any part in blaze as I switched to react now. I deleted the heroku container by mistake some time ago wasn't sure anybody is using it sorry.

MiroHibler:
Ok, guys, we're covered - @voidale was very kind to transfer the `blazejs.com` domain to me, and I also acquired `blazejs.org` which seems more appropriate now that Blaze has been transferred to community.

I've also set up Heroku app, so we only need to wait for DNS to propagate (I also redirected `blazejs.com` -> `blazejs.org`).

I'm open to any suggestion on how to proceed from now on :)

mitar:
Can you update in README links to .org version of it?

Can you make blazejs.org redirect to our GitHub pages?

https://help.github.com/articles/using-a-custom-domain-with-github-pages/

I think you will have to configure `ALIAS` DNS entry because it is a [top-level domain](https://help.github.com/articles/setting-up-an-apex-domain/).

I will configure the .org version for GitHub pages now.

mitar:
Oh, you first have to setup DNS, then I can configure it.

Can you also add me to domain to be able to manage it somehow? Which provider are you using. I can create an account there.

MiroHibler:
[...]

PR #54, opened 2016-08-23
Fix slack url

alexandesigner:
Fix the link to the Slack https://github.com/meteor/blaze/issues/51, somehow the slack spent working on different subdomains... and the badge no longer rendering. I removed to don't be a broken image.


alexandesigner:
My apologies, I realized that we're doing a redirect.

PR #65, opened 2016-08-31
Add initial site

alexandesigner:
Initial site of the blaze in this PR contains a prototype website (we still need a landing page).. but it's a start ... explore some of the material in the documentation and the guide using theme base ``hexo-theme-meteor`` the official guide Meteor;


site/source/guide.md
```diff
@@ -0,0 +1,672 @@
+---
```
mitar: This is copied from Blaze guide as it is?

site/source/index.md
```diff
@@ -0,0 +1,212 @@
+---
```
mitar: This is new or copied from somewhere?

site/source/guide.md
alexandesigner: Exactly @mitar

site/source/index.md
alexandesigner: This is basically the Readme in the package in Atmosphere ... that is what we are using in the readme too .. 

_I think we could add content to complete talking a little bit about the Blaze API ._ 😄 😄

mitar: Hm, so we will have `meteor/blaze/README.md` and `meteor/blaze/packages/blaze/README.md` (which will go to Atmosphere) and then this file. This feels to me like ...  a bit of duplication. How can we make sure that those all files are reasonable and in sync somehow?

stubailo:
After this is published, should we redirect the meteor guide page about blaze to this site?

mitar:
Yes, we could do that. We should probably then remove it from repository there as well. I would really worry to have duplicate efforts at two places.

What about Meteor documentation? And API generation? We should figure that out as well? Like how to make "link to source code" work?

stubailo:
Yeah I agree we should not duplicate content!

Let me know when we can help with that. I wrote the original JSDoc stuff but @tmeasday implemented the current build process for the API docs.

mitar:
OK, so let's start first to have automatic building and deploying. And then just keep in mind to not allow changes to Blaze related content to guide and docs but direct them here. And then we can fix links and other stuff and then link it from Meteor content.

mitar:
I will merge this for now, so that we get a version up. And then we can work on de-duplicating content. Maybe simply Atmosphere readme should be short and point to blazejs.org (because in the future it will be just one type of a package for a general Blaze package).

mitar:
BTW, I see that you have `site/source/logo/main-logo.png`, do we have to duplicate this image? Should we move also png files to `images/` and then source it from there for documentation as well?

tmeasday:
For API docs, I would suggest we move these two files
 - https://github.com/meteor/docs/blob/master/source/api/blaze.md
 - https://github.com/meteor/docs/blob/master/source/api/templates.md

Into your hexo theme, and put a pointer in the Meteor docs. You'll need to setup the hexo theme with the api-box helper and the code as a submodule.

I can do that last part if you'd prefer, just let me know where to put it.

tmeasday:
PS if you'd like to use our S3 deployment script for hexo sites, we can host the bucket in our S3 account.

mitar:
> PS if you'd like to use our S3 deployment script for hexo sites, we can host the bucket in our S3 account.

I was thinking of using GitHub pages.

tmeasday:
Sure, whatever wroks

mitar:
> Into your hexo theme, and put a pointer in the Meteor docs. You'll need to setup the hexo theme with the api-box helper and the code as a submodule.
> 
> I can do that last part if you'd prefer, just let me know where to put it.

I think that would be great if you could help. There is already an empty section for API. If you could add it there it would be great. As for files themselves, I think that for now we could add them along with other files in the docs (so `/site` directory).

mitar:
Deployed: http://blazejs.org/

PR #78, opened 2016-09-03
Resources Link Update

mrjones91:
Templates API and Blaze API links directed to the Meteor Guide which then redirects back to a page on blazejs.org. Those resource links have been updated to direct straight to the blazejs.org pages.

The Spacebars link directs to a 404 on Github and there is Spacebars package documentation on Meteor docs, so that link has been updated to the docs.

### DIFF
--- a/site/source/index.md
+++ b/site/source/index.md
@@ -198,9 +198,9 @@ Blaze will get better support for using it outside of Meteor, such as regular st
 
 # Resources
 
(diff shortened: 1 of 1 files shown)


mitar:
Oh, Spacebars documentation should probably be also moved to blazejs.org. I opened: #79

mitar:
Thanks!

But is anyway old content. README is probably the most recent content of that index.md. There is too much duplication at the moment.

PR #87, opened 2016-09-08
Update links

alexandesigner:
Updated the links in the content of the documentation and Readme

### DIFF
--- a/site/source/api/templates.md
+++ b/site/source/api/templates.md
@@ -17,8 +17,7 @@ or replaced and should be cleaned up.  You can associate data with a
 template instance, and you can access its DOM nodes when it is in the
 document.
 
-Read more about templates and how to use them in the [Spacebars](#pkg_spacebars)
-package README and the [Blaze](http://guide.meteor.com/blaze.html) article in the Meteor Guide.
+Read more about templates and how to use them in the [Spacebars](http://blazejs.org/api/spacebars) and [Blaze](http://blazejs.org/api/blaze).
 
 <h2 title="Template.<em>myTemplate</em>" id="template_myTemplate">Template Declarations</h2>
 
@@ -66,7 +65,7 @@ Then you can call this helper from template like this:
 ```
 
 You can learn more about arguments to helpers in [Spacebars
-Readme](http://blazejs.org/spacebars).
+Readme](http://blazejs.org/api/spacebars).
 
 Under the hood, each helper starts a new
 [`Tracker.autorun`](#/full/tracker_autorun).  When its reactive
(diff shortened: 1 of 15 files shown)


site/source/api/templates.md
```diff
@@ -17,8 +17,7 @@ or replaced and should be cleaned up.  You can associate data with a
 template instance, and you can access its DOM nodes when it is in the
 document.
 
-Read more about templates and how to use them in the [Spacebars](#pkg_spacebars)
-package README and the [Blaze](http://guide.meteor.com/blaze.html) article in the Meteor Guide.
+Read more about templates and how to use them in the [Spacebars](http://blazejs.org/api/spacebars) and [Blaze](http://blazejs.org/api/blaze).
```
mitar: Should we hard-code URLs or should we just have absolute URLs like `/api/spacebars` here? Because it is easier then to work locally on this content.

Not sure if Hexo has some way to specify URL placeholders which it then resolves as it is compiling markdown?

stubailo: I think you can do `../spacebars` or something like that.

mitar:
[Hexo](https://hexo.io/docs/helpers.html#URL) has some helpers. I think we should use them.

alexandesigner:
You're right, we should keep things automated.. I'm working on it!

mitar:
@alexandesigner: How is this going? I would like to release the first set of packages after this gets in. (So that READMEs on Atmosphere have updated content.) Could you look into finishing this pull request?

alexandesigner:
Wow.. that's good !

I updated the Readme in the packages ``spacebars`` ``spacebars-compiler`` ``blaze`` ``templating``, as well as the links within the content of the documentation, however, we have to see a way to update the links that are generated by the ``apibox`` (API). 

In addition, maybe it would be interesting after you make this change, complement some content related to "simple installation and use" of the package.. for example:

To install, just type

``meteor add blaze`` ...

mitar:
So: do you think it is ready for merging? Do you think something has to be done before release? Do you need any help? With fixing links from apibox? Can you be more precise what is needed there and maybe @tmeasday can help.

alexandesigner:
Yes, it's ready! Ignore the issue of apibox, the links have been updated after the site update...

>Do you think something has to be done before release?

What I think you need is just a paragraph explaining how he can install package ... 

> To install, just type
``meteor add blaze``

mitar:
> What I think you need is just a paragraph explaining how he can install package ... 

I would wait with this because I am not 100% that it will work. Especially because first the current devel version of Meteor has to be released to decouple Blaze from Meteor.

alexandesigner:
Fine, then you can do the merge.

Issue #45, opened 2016-05-20
Use Incremental DOM


lorensr:
In the hopes of improving performance. Library from google: https://github.com/google/incremental-dom

From @mitar: "So I think you want to override [materializer](https://github.com/meteor/blaze/blob/master/packages/blaze/materializer.js). If incremental DOM has good API, this might be relatively straightforward to do."

Migrated from https://github.com/peerlibrary/meteor-blaze-components/issues/120

ramezrafla:
@lorensr and @mitar, following up. Where are we and how can we help?

mitar:
So I think this could be simply something which overrides default materializer for Blaze. Maybe you should look into how current DOM materializer works and see if you can create one based on incremental DOM.

ramezrafla:
Ok - will start looking at it. Do we have somewhere where we are managing tasks? Trello maybe?

mitar:
We are using this issue tracker for tasks for now.

alexandesigner:
any progress?

aadamsx:
I asked @ramezrafla on the forums about this this other day without response.  This is one of the most critical action items IMO.

ramezrafla:
Sorry about this guys, was overwhelmed with work, we started but had to pause for urgent activities. I can get to it next week if that's ok with y'all

ramezrafla:
We started looking into the implementation of idom for Blaze, and it seems that quite a bit of work has been done with handlebars/mustache like templates: [see here](https://gist.github.com/sparhami/197f3b947712998639eb)

Looking at materializer, it seems we need to patch other libraries too, not sure yet as I am trying to figure it out. Essentially, we need to locate the root element that holds the Blaze template so we can apply idom patch to it. **Any thoughts on this would help speed up the process** (maybe we shouldn't even have materializer anymore).

mitar:
Sadly I do not have much insight here.

Maybe incremental DOM is not suitable for Blaze. One thing I was thinking is that we could simply cache HTMLJS trees after we materialize them, and then at next materialization we would just compare differences and materialize that. (Not sure if Blaze is already doing that.) That would be then very similar to React.

ramezrafla:
Ok, thanks. 

As an FYI, Incremental DOM API is actually quite simple. It can even   handle elements that we know in advance are static so it speeds up rendering.

I did some tests of sample pages with a lot of changes on an old iPhone and it is super fast. 

[...]

Issue #75, opened 2016-09-02
List of potential changes to Blaze


mitar:
Just a list of some ideas I have what we could change in Blaze to make it a bit better, keep the good parts and move on. Some of them might be breakable so we should see how to integrate this. Personally, I would like to see that we do just incremental changes. One thing to do that would be to introduce components. Then `<template>`s could be seen simply as a HTML fragments with placeholders you can reuse, without any logic. This is what I also started with Blaze Components.

But the new components (in contrast with Blaze Components) would have some changes. I will be making it a lit of comments for each of those so that community can upvote/downvote them.

Please write only ideas (and variations of existing ideas). If you want to discuss details of an idea implementation, process, roadmap, open a ticket for that particular idea (we can update then its comment here with a link to the ticket).

What are your ideas? What would you like to improve in Blaze?

mitar:
You would have to explicitly declare which arguments a component is receiving (as a dict).

mitar:
No implicit passing of a data context.

stubailo:
I'd add inline event binding like react has, and get rid of the event map style.

mitar:
The values of the arguments would be reactive, but not the whole data context (so we get immediately one level deeper equality comparison, instead of top-level data context which changes always).

mitar:
You bind events inside a template (like Blaze Components), not using event maps (they should still be supported for trickier things).

mitar:
We could offer easy way to cache results of reruns of reactivity and if it is the same, we do not have to continue.

mitar:
We could be doing diff between new and old version of internal HTMLJS and then apply just changes (like virtual DOM), maybe this is already done, not sure.

mitar:
Or use something like incremental DOM.

mitar:
For that to work, we introduce a "pure" component which does not allow any mingling with jQuery inside (no $ selector available) and all changes are result of just result of helpers and stuff.

mitar:
This is allows most things to do, the only reason I have to mingle with DOM is animations, so we should provide some way to do that.

mitar:
Arguments to methods can still be list of arguments of keyword arguments, but keyword arguments get wrapped into a normal JavaScript object which is an instance of Spacebars.kw or something, so that it looks like `options` object or something, you do not have to care or you do about the fact that it is an instance of that prototype.

mitar:
You get this object only when somebody really passed keyword arguments to your helper.

mitar:
[...]

Issue #68, opened 2016-09-01
Unnecessary re-rendering in Blaze because with's ReactiveVar uses default equals function


mitar:
(Migrated from [#4073](https://github.com/meteor/meteor/issues/4073).)

Blaze uses in its [Blaze.With](https://github.com/meteor/meteor/blob/devel/packages/blaze/builtins.js#L16) view `ReactiveVar` with default equality function:

> A function of two arguments, called on the old value and the new value whenever the ReactiveVar is set.  If it returns true, no set is performed.  If omitted, the default `equalsFunc` returns true if its arguments are `===` and are of type number, boolean, string, undefined, or null.

This means that for most data contexts which are objects re-rendering is triggered because equality function returns false. Even if new and old objects are referentially or content-wise completely equal.

The issue is that this triggers again and again `Template.currentData()` which might be in autoruns which do complex code.

So the questions are: why not provide some configurability here? Somehow one-size fits all approach is not good enough here. Some objects are equal when they are referentially equal. Some objects are equal when `EJSON.equals` returns true and it is cheaper for equality then recomputing the autorun reactive var is used in. For some you want fast failure like it is now.

Maybe a set of utilities around reactive dependencies would be beneficial which would help? https://github.com/peerlibrary/meteor-computed-field does not help with `Template.currentData()` because it does not use Blaze autorun.

In any case, the main issue is that internally all with data contexts are rerendered. And if this triggers then dynamic template inclusion to rerun, then you have things rebuilding and re-rendering all over the place.

mitar:
My current workaround is:

```
@currentData = new ReactiveVar null, EJSON.equals
@autorun (computation) =>
  @currentData.set Template.getData()
```

Probably this is what you meant as well. :-)

mitar:
I think we should address this in the following way: We add to templates API a method which allows  you to configure which equality function you want to use for data context. And this is then used for internal reactive var equality function.

So something like:

```javascript
template._dataContextEquality(EJSON.equals);
```

For now I would do it per template, but we could also try to expose some global-level function which changes a default. But I think that would be really hard to implement correctly and assure it continues to work. Personally, I think this per-template should be enough and then code which handles boilerplate (like Blaze Components) can set it for all templates.

MiroHibler:
[OT] Can we agree on using ES2015 (ES6) instead of CoffeeScript? ;)

mitar:
I will try to refrain myself. But you really do not know what you are missing. :-) It is still so much better language.

[...]

PR #103, opened 2016-09-19
Feature: improved diffing

ramezrafla:
See [Issue #45](https://github.com/meteor/blaze/issues/45)

Summary: the diff-ing engine used by Meteor lets through content that has not changed, especially relevant when we are display arrays. We added a fast comparison function and used it in two key locations to prevent needless changes to the DOM.


ramezrafla:
The tests failed as it is unable to download Meteor properly grrr .. @mitar, could you handle this at your earliest. I'd like to make an announcement on the Meteor forums as it will give an impetus to Blaze that it is getting some serious improvement in performance.

mitar:
So, this is in fact more related to the #68. It is known that comparison function can be improved, but it is unclear what is the good one for every case. Because there is a trade-off between doing comparisons (you need CPU cycles, especially on large objects, which can even potentially have cycles) and not, but then doing diffing at some other layer.

For example, doing diffing on data layer can be potentially much larger than doing diffing on DOM layer because results DOM often contains only a subset of initial data.

What exactly is the right approach is a question. For example, some projects are doing completely different thing, where they use reference equality for everything and have immutable data. So if data is immutable and reference is equal (very quick comparison), then you know that data is the same. If references are not equal, you see them as different data. Again, the idea here is that comparison on data are expensive so it might be better to do diffing on some other layer.

Blaze by default is pretty conservative here. It compares only primitive values and the rest leave to DOM layer (middle ground between deep comparison and reference comparison). #45 talks about the idea of improving the DOM layer. And #68 about the idea of improving the equality function.

What I think we should do in the big plan is (#75):
* extend Blaze templates to be classes so it is easy to extend them
* move equality function (or even data context creation function) to a method of this class, so that subclasses can override it
* make it easy to use a subclass (`<template is="subclass">`)
* even better, stop using data context, but allow custom reactive fields on the template instance, so that you can for each field configure what equality you want
* in addition to that, allow different materializes based on templates
 * some which allow 3rd party modifications (jQuery)
 * some which do virtual/incremental DOM, but do not allow 3rd party modifications
 * some which are good for animations (it does not have to be independent category)

This would allow developers to tailor their Blaze use to their particular need. Maybe their optimization point is really data, so they can to diffing there. Maybe it is DOM, because it is too much data to diff. Or too many reactivity things which get triggered anyway.

BTW, besides diffing there is also caching. I use [computed field](https://github.com/peerlibrary/meteor-computed-field) for that. The issue is that if you have three reactive sources inside autorun and one of them changes, both of those three reactive sources are invalidated and recomputed, instead of two of them returning simply the cached previous value. This is again something I think developers could then do at any granularity they want with the above changes.

I think the current Blaze default is a good conservative equality. But I think we should approach the question of equality in a more general way than trying to modify the default to one or the other special cases.

mitar:
Moreover, if you are observing issues with how arrays are rendered inside `#each`, then you might want to check/improve [this package](https://github.com/meteor/meteor/tree/devel/packages/observe-sequence).

mitar:
And a lot of these optimizations you can already do by not using data context to pass data around. So data context does not (yet) allow to configure its equality function.

ramezrafla:
> So, this is in fact more related to the #68. It is known that comparison function can be improved, but it is unclear what is the good one for every case. Because there is a trade-off between doing comparisons (you need CPU cycles, especially on large objects, which can even potentially have cycles) and not, but then doing diffing at some other layer.

So here is the challenge. With the current approach, the 'changed' objects trickle down to be compared with what is actually in the DOM, **field by field**. The comparison is far from perfect and converts nodes to DOM elements to compare them.

So the current approach is actually very heavy on CPU cycles. Using a comparison function that does it at the object level is much faster (which may partly explain why I am seeing improvements in performance)

ramezrafla:
@mitar,

I am going to raise a concern here. You have very ambitious plans for Blaze (e.g. different materializers). Who is going to maintain all this. I **strongly** recommend we keep it simple-stupid (KISS principle) and focus on improving it, and adding higher level features useful to developers that you listed in the guide. This is what is needed to **keep Blaze alive**. If we have developer time and interest, we can go deeper into Blaze and start making improvements and allow overriding etc.

ramezrafla:
Yes observe-sequence is where we need to work, except it looks like a big task at the moment (diff-ing engine). Will require more dedication. Not to mention it's a core package and we have to work with MDG developers who are super busy. Won't happen right away.

arist0tl3:
@ramezrafla Out of curiosity, I'm wondering how you're measuring/seeing performance improvements? Benchmarking?

ramezrafla:
No, purely visual. Lagging on mobile then smooth scrolling on large list (same on desktop, but less noticeable unless you have a very large list).

I knew that if I made changes to an element it would freeze a bit, that no longer happens. I know it's not scientific, but technically, we can see the problem when tracing the code and observing results.

ramezrafla:
@mitar, what is the status on this PR?

mitar:
So you made a pull request for a thing which was not discussed in advance, and which changes one of the fundamental properties of Blaze. As such you must realize that getting this change is a longer process. Moreover, as I described [on the forum](https://forums.meteor.com/t/poll-what-features-are-most-important-to-you-in-blaze-2/29521/12), it is also unclear that optimizing Blaze for one particular load improves it for all types of loads, or existing types of loads. So this can be potentially backwards incompatible change which will make some other apps slower. I have not seen any quantifiable data confirming that it works better than "it feels better on our apps" statement.

Let me quote from the forum:

> Lol, you stopped half-way through. Just because you observed optimization you think you are done? Of course is faster if you improve reactivity. But it is just partial faster from what it could be.

> Again, every code has much more data than DOM. Data >>> DOM. Because humans have to process DOM. And we are limited in how much information we can take in.

> So diffing at DOM level instead of data level means that you have to diff much much less. React observed this fact and discovered that it is faster to just compute everything and diff at the end.

> Blaze is great because it has also low-level data management. So we can have the best of both words and leave to developers to tweak things as they need. But diffing only at data level will never be as fast as React. But combining those two things, that could be the most powerful thing ever. Especially if you can tweak them to optimize for a particular use case.

> I talked with Pete Hunt (creator of React) at length about this when we organized [his talk at Berkeley](https://www.youtube.com/watch?v=MC376f3QWYw).

So I understand that this fixes your particular need now, but I am trying to look for a larger picture of Blaze. Getting something into Blaze should work for all, and should be well architectured.

This is why I asked you, before you started working on this, to look if incremental DOM/virtual DOM can be integrated with Blaze. Because that type of optimization can be done independently from other changes in Blaze. For this change, I think the order of things should be different, as [planned in the roadmap](https://github.com/meteor/blaze/milestone/4), and this is:
* we first make template extendable
* we allow developers to fine-tune what type of quality they want and need for their templates
* we see what types of loads people have and how they tuned different things to respond to those loads
* we used this to inform our decision to change defaults

So I completely agree that this is a problem. I [wrote about it](https://github.com/meteor/meteor/issues/4073) more than one year ago. What I do not believe is:
* that there is crazy urgency to change this in the next release, we lived with this for years
* there is a space for putting something in between to make this configurable, instead of just moving a knob left and right for everyone, with unclear understanding how this will influence everyone

I appreciate your work and that you investigated this well, but because it is not something which was planned, it will require also from me to look deeper into claims you are doing and try to understand them better. I also have to do some my own tests because I would like to understand better what is happening about things you are observing. So, I hear what you are saying, I am just not sure that this is the best solution to the problem you are observing. It might be, but it is still unclear to me. And it will take time to analyze this better.

Again, it is just something which is out of order and as such it requires more work.

If you want to help, you can in meantime investigate (maybe you already did) why looping in Minimongo does not work as expected. So we all agree that if one document is inserted in the middle, we would want only that DOM element to be added there and this is it. Why is this not happening? Could `#each ... in` help here? So instead of trying to mask a problem by diffing, I would like that we do not even trigger a change notification.

It would be great if we would have example instrumented app to all look into and use as a baseline.

ramezrafla:
Thanks @mitar, I was only asking as it was left in limbo on the forum https://forums.meteor.com/t/community-involvement-setup/29538/15

Essentially our goal was not incremental DOM per se, it was to improve rendering and avoid wasted calls to DOM. I don't think our solution is optimal, but it is very good with the constraints around it.

Look forward to you completing your review, in the mean time, I'll try to see if I can invest time in this:

> If you want to help, you can in meantime investigate (maybe you already did) why looping in Minimongo does not work as expected. So we all agree that if one document is inserted in the middle, we would want only that DOM element to be added there and this is it. Why is this not happening? Could #each ... in help here? So instead of trying to mask a problem by diffing, I would like that we do not even trigger a change notification.

It's likely a Tracker / diff issue (something MDG seems to have known about, and so have you)

As far as urgency, there is a message to the community: "**Blaze is getting love and it's getting better!**"

This could be a temporary fix.

mitar:
> It's likely a Tracker / diff issue (something MDG seems to have known about, and so have you)

No, this is not what is known. It is known that if a document for a particular entry changes at all, that one will be rerun completely, because default equality for data context is conservative. But this for insertion is something new, at least to me. And is not how it was documented in the past. So if this is happening, I would see this as a bug, maybe even regression. And I would like to know where it is coming from. It should work both for arrays with `_id` and Minimongo.

What I think we have to get to is:
* rerendering is triggered anytime anything in data context changes (then, we can optimize how data context changes are detected, with similar approach to yours)
* but not if some other documents in Minimongo are changed, or some other ones are inserted somewhere else, that should be fixed on Minimongo/array tracking level

> As far as urgency, there is a message to the community: "Blaze is getting love and it's getting better!"

I think nobody in Meteor community expects "be quick and break things" approach to development. In any case, I do not see that as a good reason to not do things properly. I have also not seen 1000 upvotes or comments to quickly merge on this pull request, so I do not see any urgency from others either.

I know that when you make a pull request you feel amazing and would love to get things merged. But again, when you make a pull request which was not discussed before, then you have to take responsibility to hold your own horses. This is on you.

mitar:
> With the current approach, the 'changed' objects trickle down to be compared with what is actually in the DOM, field by field. The comparison is far from perfect and converts content to DOM elements to compare them.

Yes, this is why I am proposing that we do comparison with previous value of HTMLJS for that object. So instead of converting to DOM elements, we compare HTMLJS itself. And only apply those changes which we find different between HTMLJS values.

Again, this is what I am saying that could be an interesting optimization which would work similar to virtual DOM, now that you investigated incremental DOM and are saying that it is not a good match.

And I agree that we should also optimize the whole pipeline for Tracker recomputations as well. But now imagine that we can make DOM rendering faster without Tracker recomputations optimization. Then when we have both, so great!

> You have very ambitious plans for Blaze (e.g. different materializers).

No, this is a very simple change, for this particular approach to materializing with comparing HTMLJS, because you are interested in this. Or was.

> Yes observe-sequence is where we need to work, except it looks like a big task at the moment (diff-ing engine). Will require more dedication. Not to mention it's a core package and we have to work with MDG developers who are super busy. Won't happen right away.

It would still be great if we can solve the issue there as well. I do not think we should worry too much about pull requests not getting in, they are pretty efficient on merging pull requests recently. And even if it goes late on, it goes. But somebody has to make, to even go in at all.

packages/blaze/view.js
```diff
@@ -352,13 +352,16 @@ Blaze._materializeView = function (view, parentView, _workStack, _intoArray) {
       view._isInRender = false;
 
       if (! c.firstRun) {
+        // added by ramezrafla@zegenie to skip affecting DOM for unchanged content
+        if (Blaze._deepCompare(lastHtmljs,htmljs)) {
```
mitar: Oh, you are comparing last and new HTMLJS already. Why you wrote (or I understood) that we are talking about data layer comparison.

So yea, this is similar to what I had in mind with virtual DOM. To compare old and new HTMLJS.

packages/blaze/view.js
```diff
@@ -509,21 +512,6 @@ Blaze._destroyNode = function (node) {
     Blaze._DOMBackend.Teardown.tearDownElement(node);
 };
 
-// Are the HTMLjs entities `a` and `b` the same?  We could be
-// more elaborate here but the point is to catch the most basic
-// cases.
-Blaze._isContentEqual = function (a, b) {
```
mitar: Why you didn't just improve this method?

packages/blaze/builtins.js
```diff
@@ -226,6 +226,11 @@ Blaze.Each = function (argFunc, contentFunc, elseFunc) {
         });
       },
       changedAt: function (id, newItem, oldItem, index) {
```
mitar: But this is at the data level, no?

packages/blaze/view.js
mitar: So this part of the change is something I like.

I do think we can do better though. :-)

First, why are you not simply using `EJSON.equals`? If you think that one can be optimized, than I would suggest make pull requests to it. Hm, but do we want to introduce a new dependency for Blaze?

So, one thing I think we can improve is that it seems that HTMLJS can contain not-rendered templates. See this [code snippet](https://github.com/meteor/blaze/blob/master/packages/blaze/materializer.js#L77). This means that if you are comparing any such `lastHtmljs` and `htmljs` they will not be equal, yes? So maybe it would be better if we make one pass over the HTMLJS tree, converting it into a pure HTMLJS tree. And then compare. The issue with that is that this might confuse how `rendered` callback is called. We would have to mark which segments of HTMLJS require their view 'rendered' to be called. But my worry is that now with any more complicated HTMLJS segment we are not improving much. What do you think?

Next idea: now we are all or nothing. We compare HTMLJS, if it is equal, great. But if there is one attribute difference, we still go in and then create DOM elements for whole HTMLJS tree and compare them, no, inserting/updating them? Can you point me to where DOM elements are compared?

mitar: So currently I worry that this prevents only the innermost HTMLJS trees from being rerendered.

mitar:
Can you check difference [for the 01-dbmon](https://github.com/mitar/reactconf-2015-HYPE) example between current Blaze version and one with this pull request?

packages/blaze/view.js
mitar: So I looked into this change [using this example](https://github.com/mitar/test-blaze-array) and I could not get it to in fact come to this `if (! c.firstRun) {` condition with any other view than `lookup:something`.

If I did:

```
Collection.update({}, {$set: {value: 'a'}})
```

Change was from `aa` to `a0`.

And then if I did:

```
Collection.update({}, {$set: {value: 'a', test: 'b'}})
```

Change was from `aa` to `aa`.

So, I think that `Blaze._isContentEqual` moving up is a great idea for cases like above, where I am not minimizing my fields selector well in Minimongo. But I am not sure if there is any benefit of changing the function to deep-comparison? Can you please adapt my example to the case where you get something else than those simple values `Blaze._isContentEqual` is already checking? I could imagine only very extreme cases where you would be using `{{value}}` where `value` is not a simple variable, but an object, and you are using its stringified value to be shown in the template. But is anyone really doing that?

packages/blaze/builtins.js
mitar: In [my example](https://github.com/mitar/test-blaze-array) I could not get to this issue. If I instrument `changedAt` method to print out every time is called, and then I click "add 5" button, this is not called at all, not for 5 or any other value in the Minimongo database.

mitar: I tried even to limit my fields selector to `{value: 1, test: 1}`. Then:

```
Collection.update({}, {$set: {value: 'a'}})
newItem, oldItem: Object {value: "a", _id: "s84f5D5bwFgPzwvmk"} Object {value: 0, _id: "s84f5D5bwFgPzwvmk"}
Collection.update({}, {$set: {value: 'a', test: 'b'}})
newItem, oldItem: Object {value: "a", _id: "s84f5D5bwFgPzwvmk", test: "b"} Object {value: "a", _id: "s84f5D5bwFgPzwvmk"}
Collection.update({}, {$set: {value: 'a', test: 'b', test2: 'c'}})
```

The last change changes a field outside of the projection. And `changedAt` is not called.

So, I could not reproduce that `changedAt` would be called unnecessary if a value was not really changed.

packages/blaze/view.js
ramezrafla: I could have simply renamed the new function as ._isContentEqual, except this function is meant for DOM comparison, and we are comparing objects / primitives, so to avoid confusion I broke with the past.

packages/blaze/builtins.js
ramezrafla: Yes, we have two additional diff levels right (see my original PR comment)
1. Data level
2. Field level in htmljs

I think 1 will being gone once we improve the original diff-ing which calls the observe handlers

packages/blaze/view.js
ramezrafla: Right, we had deep variables. I think for a complex App, it is to be expected. 

Here is your example adapted to our test case:
``
var obj = { a: { b : 1}};
Collection.update({}, {$set: obj});
``

packages/blaze/builtins.js
ramezrafla: Right, our test case was as per above comment:
`` var obj = { a: { b : 1}}; Collection.update({}, {$set: obj}); ``

packages/blaze/view.js
ramezrafla: > Can you point me to where DOM elements are compared?

In the past, the _isContentEqual did just that. We are no longer doing that now, we are comparing data before it becomes DOM.

> So currently I worry that this prevents only the innermost HTMLJS trees from being rerendered.

If the deepcompare finds the objects to be the same, it makes sense to skip re-rendering, unless I missed something.

mitar:
(I wanted to make this an inline reply to a inline comment, but I am realizing it is pretty valuable, so I am moving it top-level.)

Yes, sorry about that. I understood something else under "data level". For me this is more data context level, which has its own issues. Sorry for misunderstanding.

Maybe we should define terminology here better. To my knowledge we have issues at:
* data context level, equality being what it is, which triggers reruns on any non-primitive value change:
  * possible solutions:
    * get rid of data context
    * allow custom data context with custom equality function
* tracker level, which is similar to data context level, but means that in general any reactive dependency might rerun more often than necessary
 * possible solutions:
   * using something like [computed field](https://github.com/peerlibrary/meteor-computed-field) to be precise about dependency changes
* no tracker caching level: if you have autorun with multiple dependencies, when one of those change, also other are recomputed, instead of previously returned value just cached and returned
* minimongo and array processing level: we want changes to be triggered only when data really changes, and only for those changes
* blaze data level (what you are talking about here, I think): that propagation of data through blaze might be ineffective and be done when data has not really changed
* HTMLJS materialization level: when HTMLJS does not really change, but Blaze still builds DOM elements and does diff against DOM
* interacting with DOM level: querying DOM for state is often expensive
 * for more information:
  * http://kellegous.com/j/2013/01/26/layout-performance/
  * https://www.sitepoint.com/incredibly-fast-ui-animation-using-velocity-js/
  * https://davidwalsh.name/css-js-animation
 * possible solutions:
  * ignore 3rd party changes in DOM and just trust the previous HTMLJS state and apply it blindly (should probably be configurable per template)
  * do rendering only inside the `requestAnimationFrame` when DOM is frozen and do not do layout trashing, see [example](https://github.com/DispatchMe/meteor-kernel)

Did I miss anything?

ramezrafla:
@mitar 
You obviously know a lot more of the Blaze history than I do, I just jumped recently to resolve the needless re-rendering when no data has changed. I am sure I will learn all these issues as I go carefully through the code and the issues.

> ignore 3rd party changes in DOM and just trust the previous HTMLJS state and apply it blindly (should probably be configurable per template)

We have an App that affects the DOM directly and had no issues. I think it has to do with carefully managing collisions. We do not with JS affect anything that is handled by Blaze. If that separation is done, we're good

> blaze data level (what you are talking about here, I think): that propagation of data through blaze might be ineffective and be done when data has not really changed
HTMLJS materialization level: when HTMLJS does not really change, but Blaze still builds DOM elements and does diff against DOM

Right! So the first one has to do with the diff-ing algorithm in the compare-sequence lib you had talked about before. This is where the real fix should be applied, unless it breaks something else (is this used by Tracker?)

The second is where I believe we have a lot of savings. We don't build DOM elements to diff against DOM. This is bad design. We should always diff raw data, not display layer. This is what makes Blaze fast (like VirtualDOM) and is the reason Incremental DOM is slower than Virtual DOM.

packages/blaze/view.js
mitar: No, it does not happen. So what I am saying is that you will never really do in your app:

```
{{value}}
```

Where:

```
value = {a: {b: 1}};
```

You might do:

```
{{value.a.b}}
```

But what I am saying is that in the comparison, Blaze just compares then values of `b`s, so primitive values.

So can you give me a realistic example where the value of `lastHtmljs` and `htmljs` is something else than what it is checked by `Blaze._isContentEqual`? I have [updated my example](https://github.com/mitar/test-blaze-array) with complex objects to test this.

packages/blaze/builtins.js
mitar: So I could not reproduce that `changedAt` when data projected using `fields` didn't really change.

For example, if I have the following cursor over which I am doing `#each`:

```
Collection.find({}, {sort: {'value.a': 1}, fields: {'value.a': 1, test: 1}});
```

And values are inserted like:

```
Collection.insert({value: {a: 1}});
```
When I do:

```
Collection.update({}, {$set: {'value.b': 'b'}})
```

`changedAt` is not called at all.

And if I insert one one document in the middle, only Blaze for that one document is called, and no `changedAt` at all, for nothing.

ramezrafla: The only difference with what I did is that it was server side update with method. If you can replicate it, it means that the problem is in the publish / diff-ing.

packages/blaze/view.js
ramezrafla: Oh, I just used deepCompare for the sake of convenience and because I haven't gone through all the options. deepCompare short circuits comparison if it's a primitive so I am reusing the same function for risk mitigation as the data level.

But I did see somewhere in the code that htmljs can be views. Also, we skipped DOM comparison altogether that was there before, which was my biggest concern. If I was too cautious, then we can simply simplify the compare function. But, why would the original developers convert to DOM if it was always primitives?

packages/blaze/builtins.js
mitar: Could you please make an example showing what you observed?

We need tests for these changes anyway.

packages/blaze/view.js
mitar: > But I did see somewhere in the code that htmljs can be views. Also, we skipped DOM comparison altogether that was there before, which was my biggest concern. If I was too cautious, then we can simply simplify the compare function.

Yes, but at a higher level. These views which have `!computation.firstRun` seems to be very simple ones.

Which is also a problem probably, that higher-levels are recreated.

So can you make in your codebase (on which you are doing these optimizations here) a check to print out when `computation.firstRun` is false and `htmljs` is something complicated.

> But, why would the original developers convert to DOM if it was always primitives?

I have no idea why they convert at all before the `if`. The fact that you moved `if` out is great, but I have no idea why they did it like this. Does converting to DOM has some side-effects we do not know?

mitar: > In the past, the _isContentEqual did just that. We are no longer doing that now, we are comparing data before it becomes DOM.

I mean, where it is comparing does it have to replace a DOM element with a new one. It seems to me that it does not do a new DOM element and then compares with current DOM, if necessary. But it removes a DOM element if previous view is invalided, it then creates a DOM element for new state of the view, and adds it. Even if comparing previous and new DOM element would show that it is almost the same, or even exactly the same.

> If the deepcompare finds the objects to be the same, it makes sense to skip re-rendering, unless I missed something.

Yes, but I see those checks to be run only for very leaf rendering, nothing before. :-(

mitar:
> You obviously know a lot more of the Blaze history than I do, I just jumped recently to resolve the needless re-rendering when no data has changed. I am sure I will learn all these issues as I go carefully through the code and the issues.

Oh, maybe I do know about history, but about internals we are all learning. I think you looked much deeper than I ever did in details of rendering itself. So we are all learning here.

> We have an App that affects the DOM directly and had no issues. I think it has to do with carefully managing collisions. We do not with JS affect anything that is handled by Blaze. If that separation is done, we're good

Yes, me too. Or you leave to Blaze, or you use something else. But do not mix. :-)

> So the first one has to do with the diff-ing algorithm in the compare-sequence lib you had talked about before. This is where the real fix should be applied, unless it breaks something else (is this used by Tracker?)

But I have not been able to reproduce it so that I could look what is happening.

But I was using Minimongo and not arrays directly.

> The second is where I believe we have a lot of savings. We don't build DOM elements to diff against DOM. This is bad design. We should always diff raw data, not display layer. This is what makes Blaze fast (like VirtualDOM) and is the reason Incremental DOM is slower than Virtual DOM.

Yes, but I could not really find where diffing is done. I have found diffing for attributes, but for DOM it seems like Blaze simply removes old element and adds a new one, when something changes. It does not try to diff and patch the old one?

ramezrafla:
From view.js
``
var rangesAndNodes = Blaze._materializeDOM(htmljs, [], view);
if (! Blaze._isContentEqual(lastHtmljs, htmljs)) {
domrange.setMembers(rangesAndNodes);
Blaze._fireCallbacks(view, 'rendered');
}
``
So this creates the DOM elements before comparing

mitar:
> So this creates the DOM elements before comparing

Yes, but it compares HTMLJS, not DOM.

For DOM it seems (inside `setMembers`) that it simply removes old and adds new, I do not see any logic to really compare old DOM and new DOM to see the minimal change. Hm.

mitar:
BTW, this is waiting for tests/reproducible example for the `changedAt` issue.

The second change I like and I think we should add it to Blaze, but I will maybe add it with `_isContentEqual`, but also tests would be needed. This one seems pretty easy to do.

mitar:
So I think this should be split into two pull requests. One for moving `_isContentEqual`, and one for data. And we need tests for both, to have a case when that code is match at all.

mitar:

mitar:
Based on this pull request, I made a change in e17991a, which prevent unnecessary materialization of DOM. This is what I have able to make a reproduction myself. For the other part, checking a change in data layer I have not been able to make a reproduction.

I am closing this pull request. If we find a good reproduction for that other case, we could look into optimizing that as well.

For more ideas how to optimize Blaze see #111.

arggh:
@mitar Did you benchmark this change on any of your test cases?

mitar:
Sadly not, I do not have any real case where this is happening. I am relaying on the report in this pull request. Feel free to make a test case/benchmark and help here.

arggh:
I might have something useful in one of my projects for a benchmark.

ramezrafla:
Beautiful @mitar! Thanks!

@arggh, we caught this as we outputted to the console every re-render. This can improve performance dramatically by skipping DOM manipulation if we have a very dynamic page. I am guessing this change is going to be very noticeable for a lot of people. Look forward to your benchmark.

There is also room for improvement in the actual diffing but that's a bigger job.

mitar:
One more thing. I was unable to get to `not c.firstRun` for anything else besides primitives for which `Blaze._isContentEqual` is good at checking equality. If anyone can get there with something which `Blaze._isContentEqual` is not comparing well, we might want to upgrade `Blaze._isContentEqual` to something which checks values in depth.

Issue #98, opened 2016-09-14
Make Templates an extendable class


mitar:
- [ ] change Blaze templates to be nicely extendable JavaScript classes
- [ ] move resolving of `this` and any symbols to methods of a class instead of a statically defined methods on `Template`
- [ ] add life-cycle and some other methods to this base Blaze class, a bit like Blaze Components
- [ ] add a method which prepares a data context (so that we can do data context validation, or custom equality on the data context, or replace it with reactive dict, or disable it all together)
- [ ] have a method for lookup, migrated from https://github.com/meteor/meteor/issues/5577

filipenevola:
I'm closing this issue because it's too old.

If you think this issue is still relevant please open a new one.

Why? We need to pay attention to all the open items so we should keep opened only items where people are involved.

PR #116, opened 2016-09-25
Converting Template to a class

mitar:
TODO:
- [ ] update documentation and terminology everywhere

See: #98

### DIFF
--- a/packages/blaze/builtins.js
+++ b/packages/blaze/builtins.js
@@ -316,7 +316,7 @@ Blaze._TemplateWith = function (arg, contentFunc) {
     // Since we are generating the Blaze._TemplateWith view for the
     // user, set the flag on the child view.  If `content` is a template,
     // construct the View so that we can set the flag.
-    if (content instanceof Blaze.Template) {
+    if (Blaze.isTemplate(content)) {
       content = content.constructView();
     }
     if (content instanceof Blaze.View) {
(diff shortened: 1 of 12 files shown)


mitar:
Please test. :-) It should be backwards compatible if you have not done any strange things. But it would be people using Blaze extensively would test it.

mitar:
Not all tests are passing yet. Maybe wait a bit before testing it with extra code. :-)

aadamsx:
How would we test something like this on an existing project running 1.3.5.1?

mitar:
I [documented how to run Meteor with these packages here](https://github.com/meteor/blaze/blob/master/CONTRIBUTING.md). I think it should just work with 1.3.5.1 as well, but see.

If you have `Package.blaze.Blaze.Template.newRenderFunction` in the console, then you are running this branch.

arist0tl3:
So I just walked through a shorter one of our regression tests with the `template-class` branch and everything looked good, functionally.

I did notice a ton of warning logs that looked like:

`Assigning helper with Template.X.Y = ... is deprecated.  Use Template.X.helpers(...) instead.`

But, in each instance, Y wasn't actually a helper, but a nested template. So, for the following:

```
<template name="myForm">
  {{> myInput args...}}
</template>

<template name="myInput">
  <!-- stuff -->
</template>
```

I am seeing:

`Assigning helper with Template.myForm.myInput = ... is deprecated.  Use Template.myForm.helpers(...) instead.`

@mitar If you want more info, let me know as it's a private repo.

filipenevola:
Hi @mitar, do you want to have this PR merged yet?

If yes, please fix the conflict and comment here so we can re-open, review and merge.

PR #124, opened 2016-09-28
Added backers and sponsors from OpenCollective

piamancini:
now sponsors&backers will automatically appear here. An example of how it'll look https://github.com/apex/apex#backers


mitar:
Looks great! Thanks!

PR #135, opened 2016-10-12
Remove documentation leftovers

povesteam:
Probably left here when the documentation was moved from one style to another.

### DIFF
--- a/site/source/index.md
+++ b/site/source/index.md
@@ -3,7 +3,7 @@ title: Overview
 description:
 ---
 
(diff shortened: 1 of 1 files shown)


stubailo:
Good catch!

abernix:
This seems to have completely broke the generating of the docs which wasn't caught by the CircleCI deploy as a failure because this repo was missing https://github.com/meteor/hexo-s3-deploy/commit/278a408c38997b08679779baec805c44a99a78ed (PR submitted for that).

I believe `{% raw %}` and `{% endraw %}` are still required when the tag is a`{{# blah }}` statement (with the `#`, versus `{{blah}}`)?  They're still used throughout meteor/docs, at least.

Without reverting this change, the following is produced during a `hexo generate`:

```
FATAL Something's wrong. Maybe you can find the solution here: http://hexo.io/docs/troubleshooting.html
Template render error: (unknown path) [Line 1, Column 335]
  unexpected token: #
```

Anyhow, I can confirm that adding `{% raw %}` / `{% endraw %}` back fixes it.

PR #137, opened 2016-10-14
Correct anchor link for {{each .. in ..}}

xavcz:
Hey! A link was broken from spacebars to the benefits of each .. in.

ps: the new site for blaze is awesome 🎉


mitar:
Thanks!

PR #139, opened 2016-10-18
Remove redundant code/text, correct list of falsy values, edits

thoragio:
- Removed redundant code block and text since
- `[]` is not falsy whereas NaN is
- Various spacing edits and minor text edits to (hopefully) make it a bit clearer

### DIFF
--- a/site/source/index.md
+++ b/site/source/index.md
@@ -7,9 +7,9 @@ Blaze is a powerful library for creating user interfaces by writing reactive HTM
 
 ## Quick Start
 
-Blaze is a Meteor-only package for now. Sooner we will have Blaze on npm so you can use it in your stack.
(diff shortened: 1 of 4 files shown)


site/source/index.md
```diff
@@ -54,33 +54,6 @@ Here are two Spacebars templates from an example app called "Leaderboard" which
 The template tags `{{name}}` and `{{score}}` refer to properties of the data context (the current player), while `players` and `selected` refer to helper functions.  Helper functions and event handlers are defined in JavaScript:
 
 
-```html
```
mitar: So why removing this example?

thoragio: Because the code example and that paragraph of text is shown twice (one right after the other). Take a look:
https://github.com/meteor/blaze/blob/master/site/source/index.md

mitar:
Thanks.

PR #161, opened 2016-11-15
Add back raw/endraw to `{{#if}}/`{{#each}}` in intro to docs

abernix:
meteor/blaze#135 seems to have introduced a change which broke `hexo generate` and no doc deployment has occurred since it was merged.

For example: #139 is not live and thus the duplicated content (which is what caused me to start researching this originally) is still on blazejs.org.  #137 is also not deployed.

More comments here:  https://github.com/meteor/blaze/pull/135#issuecomment-260718442:

`{% raw %}` and `{% endraw %}` are still required when the tag is a `{{# blah }}` statement (with the `#`, versus `{{blah}}`)

This PR simply adds them back.

### DIFF
--- a/site/source/index.md
+++ b/site/source/index.md
@@ -3,7 +3,7 @@ title: Overview
 description:
 ---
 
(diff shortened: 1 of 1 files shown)


mitar:
Thanks!

PR #227, opened 2017-01-05
Removes jQuery dependency

marxo:
Removes the `$.trim()` dependency of jQuery from token parsing. Swaps jQuery for regex trim function.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -7,6 +7,7 @@
   Based on [#50](https://github.com/meteor/blaze/pull/50).
 * Prevent unnecessary materialization of DOM.
   Based on [#103](https://github.com/meteor/blaze/pull/103).
+* Removes a jQuery dependency on $.trim() in `attrs.js#L175`
 
 ## v2.2.1, 2016-Dec-31
 
--- a/packages/blaze/attrs.js
+++ b/packages/blaze/attrs.js
@@ -171,9 +171,8 @@ var StyleHandler = Blaze._DiffingAttributeHandler.extend({
         tokens.remove(match[1]);
       }
 
-      // XXX No `String.trim` on Safari 4. Swap out $.trim if we want to
-      // remove strong dep on jquery.
-      tokens.append(match[1], match[0].trim ? match[0].trim() : $.trim(match[0]));
+      // XXX No `String.trim` on Safari 4. Swap out for manual trim regex
+      tokens.append(match[1], match[0].trim ? match[0].trim() : match[0].replace(/^\s+|\s+$/g, ''));
 
       match = regex.exec(attrString);
     }


marxo:
@mitar For several reasons:
1. It's a gateway bug that will get me to contribute more, I wanted to have a feeling of the contrib process.
2. It can only make things better.
3. If it only makes things better there's no reason not to start with it.

mitar:
But why is this a bug? It does not work otherwise?

marxo:
@mitar It's not a bug, just a small improvement - it will not use jQuery anymore to shim the trim.

mitar:
Any why is this an improvement?

marxo:
@mitar "It will not use jQuery anymore."

mitar:
I think you will have to learn to contribute a bit better if you want to contribute. This is not how contributing works. I suggest you look at issues with label [contributions welcome](https://github.com/meteor/blaze/issues?q=is%3Aissue+is%3Aopen+label%3A%22contributions+welcome%22) and start with those. Those are ideas which already have some agreement that they are necessary.

For unwarranted pull requests you have to have a bit more discussion and support why is this really necessary and the burden of proof is on you. Often it is better to first open an issue explaining your idea so that you can discuss it with others before you start working on it. For example, in this context you could open issue like "Remove jQuery dependency" and explain why it would be so important to remove jQuery dependency from Blaze.

What would happen in that discussion would probably be that you would learn that Blaze is tightly linked to jQuery and it is not possible to remove it as a dependency without breaking backwards compatibility. And because we care a lot about backwards compatibility, you would really have to explain how to solve this issue.

And because jQuery will be hard to remove as a whole, it is unclear why removing it in this particular instance is important?

msavin:
My understanding is we want to make Blaze work without jQuery - and/or to make jQuery optional - so this looks to be a welcomed PR

mitar:
Why would we want that for Blaze 1? How would you implement `template.$` without jQuery? And event processing.

sbr464:
I agree that it would be welcome to remove jquery but there would be other
components needing solutions, as mentioned by .$ selectors/events etc. I
think there are potential benefits of using blaze and react together when
needed but also having to pull in jquery by default makes this less
desirable. If there are non user facing / blaze internals that could do
without jquery easily, I vote to nix them.

sbr464:
Is there a place to discuss these things for blaze 2 currently? Or is that
not currently a formal discussion?

mitar:
For Blaze 2 there are multiple issues discussing various possible directions. Nothing is yet clear about that. I think the best for this particular issue would be to open a new issue about removing jQuery dependency and then somebody should do a good investigation on all aspects of how Blaze depends on jQuery. I remember at least multiple (not completely sure about them): use of jQuery data context, use of jQuery event delegation and event system in general, use of `template.$`.

PR #236, opened 2017-02-26
Updated package.js for templating-tools and spacebars-compiler

sethmurphy18:
Made changes suggested by @abernix and @mitar. Also updated `package.js` for `templating-tools` to accept the new version of `minifier-js`, since the package never uses the API directly.

### DIFF
--- a/packages/spacebars-compiler/package.js
+++ b/packages/spacebars-compiler/package.js
@@ -5,7 +5,6 @@ Package.describe({
   git: 'https://github.com/meteor/blaze.git'
 });
 
-// Use uglify-js from NPM
 Npm.depends({
   'uglify-js': '2.7.5'
 });
--- a/packages/spacebars-compiler/package.js
+++ b/packages/spacebars-compiler/package.js
@@ -5,6 +5,11 @@ Package.describe({
   git: 'https://github.com/meteor/blaze.git'
 });
 
+// Use uglify-js from NPM
+Npm.depends({
+  'uglify-js': '2.7.5'
+});
+
 Package.onUse(function (api) {
   api.use('underscore@1.0.9');
 
@@ -21,11 +26,6 @@ Package.onUse(function (api) {
     'codegen.js',
     'compiler.js'
   ]);
-
-  // Pull in uglify-js from NPM
-  Npm.depends({
-    'uglify-js': '2.7.5'
-  });
 });
 
 Package.onTest(function (api) {
@@ -47,8 +47,4 @@ Package.onTest(function (api) {
     'compile_tests.js',
     'compiler_output_tests.coffee'
   ]);
-
-  Npm.depends({
-    'uglify-js': '2.7.5'
-  })
 });
--- a/packages/templating-tools/package.js
+++ b/packages/templating-tools/package.js
@@ -16,7 +16,7 @@ Package.onUse(function(api) {
     // boilerplate-generator uses spacebars-compiler.)
     // XXX maybe uglify should be applied by this plugin instead of via magic
     // weak dependency.
-    'minifier-js@1.2.14 || =2.0.0'
+    'minifier-js@1.2.14 || 2.0.0'
   ]);
 
   api.export('TemplatingTools');
--- a/packages/templating-tools/package.js
+++ b/packages/templating-tools/package.js
@@ -8,15 +8,7 @@ Package.describe({
 Package.onUse(function(api) {
   api.use([
     'underscore@1.0.9',
-    'ecmascript@0.5.8',
-
-    // minifier-js is a weak dependency of spacebars-compiler; adding it here
-    // ensures that the output is minified.  (Having it as a weak dependency means
-    // that we don't ship uglify etc with built apps just because
-    // boilerplate-generator uses spacebars-compiler.)
-    // XXX maybe uglify should be applied by this plugin instead of via magic
-    // weak dependency.
-    'minifier-js@1.2.14 || 2.0.0'
+    'ecmascript@0.5.8'
   ]);
 
   api.export('TemplatingTools');
--- a/packages/templating-tools/package.js
+++ b/packages/templating-tools/package.js
@@ -16,7 +16,7 @@ Package.onUse(function(api) {
     // boilerplate-generator uses spacebars-compiler.)
     // XXX maybe uglify should be applied by this plugin instead of via magic
     // weak dependency.
-    'minifier-js@1.2.14'
+    'minifier-js@1.2.14 || @2.0.0'
   ]);
 
   api.export('TemplatingTools');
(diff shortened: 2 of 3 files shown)


sethmurphy18:
Alirght, @mitar all done, PR is done and ready to merge.

packages/spacebars-compiler/package.js
```diff
@@ -5,6 +5,11 @@ Package.describe({
   git: 'https://github.com/meteor/blaze.git'
 });
 
+// Use uglify-js from NPM
```
mitar: I mean, this comment is unnecessary. It is pretty clear that this is what this code does.

packages/templating-tools/package.js
```diff
@@ -16,7 +16,7 @@ Package.onUse(function(api) {
     // boilerplate-generator uses spacebars-compiler.)
     // XXX maybe uglify should be applied by this plugin instead of via magic
     // weak dependency.
-    'minifier-js@1.2.14'
+    'minifier-js@1.2.14 || 2.0.0'
```
mitar: @abernix, is this good?

abernix: Hmm, this does seem to be defeating the purpose of the original change.  Did I miss the discussion leading to this change somewhere?

@sethmurphy18 The original intention of #234 was to get [this change](https://github.com/meteor/blaze/pull/234/commits/d2ee1ac5ca2f9e7b943699e31924b316d1cbc954) included because the new minifier would actually minify in a different way.

Looks like that was reverted with https://github.com/meteor/blaze/pull/234/commits/5594d8095c81ea1bb09d66914814a5e725d012ff  and that possibly the only reason that Blaze is passing this is because we're explicitly pinning it to the current `1.2.14` version via this line (`2.0.0` is not published yet)?

I understand the changes to `spacebars-compiler` in order to get `_beautify` working and I think the direct dependency on the `uglify-js` NPM is fine for that, but what happened to the `templating-tools` change to compensate for the Babili minification?  Don't we still need that when the official `minifier-js` (the `2.0.0` in this version constraint) is published?

sethmurphy18: This issue lies in Meteor will only install 1 version of each package at a time. Since we are bumping the version of `minifier-js` to `2.0.0`, meteor will fail to build because there is no package installed that matches those contraints because `@1.2.14` means anything 1.x.y where x is >= 2 and y >= 14, but that first number has to be 1. The new contraints I added, say x.y.z, where x = 1 or 2, if x is 1 refer to above. If x is 2, y and z must be >= 0. Ideally, we would just leave the version off, but according to the docs that is bad practice.

sethmurphy18: To be 100% honest, I am not sure why the package is even included. The minifier isn't accessed anywhere in the code. I am not sure if isobuild will minify code from something just because it uses a minifier plugin.

packages/spacebars-compiler/package.js
sethmurphy18: I will remove it and commit once we figure out the second issue here.

packages/templating-tools/package.js
sethmurphy18: I just finished running a search through the entire Blaze code, the minify function from the original `minifier-js` package isn't used anywhere. The reason the tests were failing originally is because the beautification part wasn't there, which was an issue with `spacebars-compiler`. The reason that `templating-tools` was failing is because while `spacebars-compiler` had failsafes in place in case it wasn't there, `templating-tools` did not. In my opinion that was the only reason the package was included as a strong dependency in `templating-tools` to begin with. It had nothing to do with minification, it was beautification.

sethmurphy18: Search results for validation:

```
Searching 11681 files for "UglifyJSMinify"

/home/seth/Projects/blaze/packages/spacebars-compiler/compile_tests.js:
   21          // remove initial and trailing parens
   22          string = string.replace(/^\(([\S\s]*)\)$/, '$1');
   23:         if (! (Package['minifier-js'] && Package['minifier-js'].UglifyJSMinify)) {
   24            // these tests work a lot better with access to beautification,
   25            // but let's at least do some sort of test without it.

/home/seth/Projects/blaze/packages/spacebars-compiler/compiler.js:
    1: var UglifyJSMinify = Npm.require('uglify-js').minify;
    2  
    3  SpacebarsCompiler.parse = function (input) {
    .
   97  
   98  SpacebarsCompiler._beautify = function (code) {
   99:   var result = UglifyJSMinify(code, { 
  100      fromString: true,
  101      mangle: false,

```

sethmurphy18:
Also going to bump the revision of `spacebars-compiler` due to a change that isn't an API change. I forgot to do that on the initial PR.

sethmurphy18:
I'm also going to bump the revision of `spacebars-compiler` since I made a change that didn't affect that API. I forgot to do that on my last PR.

packages/templating-tools/package.js
sethmurphy18: Ok, I tried removing it and commited to a [test branch](https://github.com/eagerestwolf/blaze/tree/test-minifier-js) and I am running a [Circle build](https://circleci.com/gh/eagerestwolf/blaze/40) now to see if everything passes or fails.

sethmurphy18: All of the tests passed even without that package. I am building meteor now to see if I get errors over there.

sethmurphy18: Ok, no errors on the meteor side, except `blaze` (still) and Mongo oplog, which sometimes passes sometimes fails (I think that one is related to my PC), and DDP which got a timeout instead of a rate-limit, which again I think is my PC. `spacebars-compiler` and `templating-tools` both pass, though, on client and server, which means to me that either `minifier-js` is not needed by `templating-tools` or it's not tested. Another thing of note is that *everything* sent to the client is concatenated and minified by default, courtesy of `isobuild`, so long as you aren't in development mode, so test and production modes will be minified by default.

sethmurphy18:
@abernix and @mitar what are both of your opinions on this PR? Should we keep `minifier-js` since it is doing nothing, or should we nix it?

mitar:
I think I would just remove dependency on `minifier-js@1.2.14` from `templating-tools` package and document in the `History.md` that one should decide on minifying by yourself for your app by adding it to the Meteor app. I think this whole dependency from `templating-tools` to force minimization was done before all this changes to minification becoming its own build step and stuff like this.

sethmurphy18:
Sounds like a plan. Nixed it will be.

sethmurphy18:
Added history entry under `vNEXT` since this version is not yet released.

packages/spacebars-compiler/package.js
mitar: Ping for this as well.

mitar:
Cool I think this looks good. I will change text in history a bit after merging, but it looks otherwise very good. Thanks!

sethmurphy18:
No problem. Sorry about that comment. Forgot to remove it, thanks for pointing it out though. Push new commit. Should be building now.

PR #248, opened 2017-04-12
DOCS: Fix broken internal links

coagmano:
I found several broken links in the Blaze docs with incorrect capitalisation or 
change from `_` to `-` and relative links to Meteor docs. I've edited links in 
JSDOC blocks and in `site` where I found issues.

Not sure how to rebuild the docs, so haven't tested a full build.

### DIFF
--- a/packages/blaze/builtins.js
+++ b/packages/blaze/builtins.js
@@ -89,7 +89,7 @@ Blaze.If = function (conditionFunc, contentFunc, elseFunc, _not) {
 };
 
 /**
- * @summary An inverted [`Blaze.If`](#blaze_if).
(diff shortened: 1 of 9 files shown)


mitar:
Thanks!

mitar:
You can see [instructions how to build here](https://github.com/meteor/blaze/tree/master/site#starting-hexo). Could you try that and see if everything works?

coagmano:
Building the docs was a lot simpler than I thought it would be haha.

I've checked all the links and found a couple of stragglers. Otherwise all good.

I found another issue with linked `dl>dt>dd` anchors not having an offset generated from the header. Don't have time to fix that now, so I'll pop that in another issue for later.

mitar:
Thanks!

PR #267, opened 2017-08-02
Remove version constraints from spacebars-compiler test dependencies.

benjamn:
These version constraints are unnecessary, since `./meteor test-packages` runs from a checkout, and all of the depended-on packages can be found in `meteor/packages`, `meteor/packages/non-core`, or `meteor/packages/non-core/blaze/packages`.

This should fix the problem noted by @GeoffreyBooth in this [comment](https://github.com/meteor/meteor/pull/8960#issuecomment-319726289).

### DIFF
--- a/packages/spacebars-compiler/package.js
+++ b/packages/spacebars-compiler/package.js
@@ -1,7 +1,7 @@
 Package.describe({
   name: 'spacebars-compiler',
   summary: "Compiler for Spacebars template language",
-  version: '1.1.2',
+  version: '1.1.3',
   git: 'https://github.com/meteor/blaze.git'
 });
 
@@ -29,14 +29,11 @@ Package.onUse(function (api) {
 
 Package.onTest(function (api) {
   api.use([
-    'underscore@1.0.9',
-    'tinytest@1.0.11',
-    'coffeescript@1.2.4'
-  ]);
-
-  api.use([
+    'underscore',
+    'tinytest',
+    'coffeescript',
     'spacebars-compiler',
-    'blaze-tools@1.0.10'
+    'blaze-tools'
   ]);
 
   api.addFiles([


benjamn:
@mitar any idea what that test failure is about?

mitar:
Tests are failing at a different place, so we can ignore that.

Are you sure this is OK? I mean, somebody from MDG told me initially that now that Blaze was moved out, we have to put all those versions in. So your explanation above is about testing Meteor repository. But these packages could also be tested outside of Meteor repository?

So maybe I am confused here how things work, but I had also some bad experience in the past with those test dependencies. So if you are saying this is OK, we can merge like this. Then this should also be done in other Blaze packages?

packages/spacebars-compiler/package.js
```diff
@@ -29,14 +29,11 @@ Package.onUse(function (api) {
...
-    'coffeescript@1.2.4'
-  ]);
-
-  api.use([
+    'underscore',
+    'tinytest',
+    'coffeescript',
     'spacebars-compiler',
-    'blaze-tools@1.0.10'
+    'blaze-tools'
```
mitar: Are you sure `blaze-tools`, a non-core package does not require a version?

benjamn: The reason this is fine is because these are `Package.onTest` dependencies, and we run package tests with a checkout version of Meteor, which has access to local copies of all these packages.

benjamn:
I'm going to merge this now, so that we can make progress on https://github.com/meteor/meteor/pull/8960. If there are any problems, don't hesitate to let me know, and/or add the versions back. The main one that caused problems was the `coffeescript` version, so I'm hoping we can avoid that constraint, at least.

mitar:
OK.

PR #268, opened 2017-08-10
Add docs for multiple event types in a single declaration

coagmano:
I was digging around the internals and found that you can add a handler for multiple event types on the same selector by separating each event with a forward-slash `/`, eg:

```
Template.foo.events({
    'click/keyup/touchend .submit' (event, inst) {
        // Do something...
    },
});
```

([Code which enables it here](https://github.com/meteor/blaze/blob/master/packages/blaze/events.js#L131))

Such a handy undocumented feature that I decided to document it

### DIFF
--- a/site/source/api/templates.md
+++ b/site/source/api/templates.md
@@ -344,7 +344,8 @@ forms:
 
 <dl>
 {% dtdd name:"<em>eventtype</em>" %}
-Matches a particular type of event, such as 'click'.
+Matches a particular type of event, such as 'click'. Can contain 
+multiple types separated by a `/` eg. 'touchend/mouseup/keyup'
 {% enddtdd %}
 
 {% dtdd name:"<em>eventtype selector</em>" %}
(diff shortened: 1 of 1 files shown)


site/source/api/templates.md
```diff
@@ -344,7 +344,8 @@ forms:
 
 <dl>
 {% dtdd name:"<em>eventtype</em>" %}
-Matches a particular type of event, such as 'click'.
+Matches a particular type of event, such as 'click'. Can contain 
+multiple types separated by a `/` eg. 'touchend/mouseup/keyup'
```
mitar: Maybe dot at the end? Make a full sentence. So, please improve punctuation here. Commas, dots, etc.

site/source/api/templates.md
```diff
@@ -377,8 +378,12 @@ Example:
   // Fires when any element with the 'accept' class is clicked
   'click .accept'(event) { ... },
 
+  // Fires when 'accept' is clicked or focused
+  'click/focus .accept'(event) { ... }
+
   // Fires when 'accept' is clicked or focused, or a key is pressed
-  'click .accept, focus .accept, keypress'(event) { ... }
```
mitar: Can you leave the existing example as well?

coagmano:
Okay I've edited it to _hopefully_ be a bit clearer, and kept the existing examples

mitar:
Great. Thanks!

PR #279, opened 2018-02-09
spacebars-compiler optimisations

nathan-muir:
Hi,

Found a few instances where the spacebars compiler aborts optimisations unnecessarily.

Case 1: Inside Block tags, eg `{{#if }} ... {{else}} ... {{/if}}`
Case 2: More than one level of nested elements.  (Compacting `HTML.Raw` fails)
Case 3: Inside `<tables>` (caused by a pessimistic workaround to auto-injection of `<tbody>` when using `innerHTML`.)

The resolution for Case 1 turned out messier than I had hoped. Due to block tags like `{{#markdown}}` that don't contain HTML. Rather than exposing the `textMode` prop, maybe something like `isHTML`?

----

My motivation for digging through this code, is to implement a `<template whitespace="strip">` mode [1]. Which would work similar to react (Strip any whitespace that contains a newline).

Early comparisons show that this saves around ~100kb (uncompressed) from my client bundle.

[1] https://github.com/3stack-software/blaze/commit/de3d4215e29d71bd973814398c223e57acab19da

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,7 +4,7 @@ checkout:
     - git submodule update --init --recursive
 machine:
   node:
-    version: 0.12
+    version: 8
   pre:
     - git config --global user.name "circleci"
     - git config --global user.email "circleci@users.noreply.github.com"
@@ -22,7 +22,7 @@ test:
     #    parallel: true
     - METEOR_COMMAND=/tmp/meteor/meteor tests/test-runner/test-all.sh:
         parallel: true
-    - cd site; npm run generate
+    - cd site; npm run build
 deployment:
   s3:
     branch: /^(master|version-.*)/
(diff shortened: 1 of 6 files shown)


nathan-muir:
Tests are passing (eg. `$ METEOR_COMMAND=/tmp/meteor/meteor tests/test-runner/test-all.sh`)
but it seems circle-ci is failing on `cd site; npm run generate`

filipenevola:
Hi @nathan-muir, do you want to have this PR merged yet?

If yes, please fix the conflict and comment here so we can re-open, review and merge.

nathan-muir:
Hi @filipenevola Can you reopen the PR. I've rebased this set of changes to resolve the conflicts.

filipenevola:
@nathan-muir please open a new PR, GitHub is not allowing me to re-open this one as the branch was rebased.

PR #284, opened 2018-04-24
Fix Regressed Test

toinevk:
This test has regressed in modern browsers due to intentional change in form action behaviour in Chrome 66 Safari 11.0 and Firefox 59. An empty or missing form action will now return the document's base URL. To receive the actual action getAttribute('action') must be called. If
HTML attribute being checked is form, check against getAttribute('action') instead.
https://bugs.chromium.org/p/chromium/issues/detail?id=724596

Also safe for legacy browsers: https://caniuse.com/#search=getAttribute

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,7 +4,7 @@ checkout:
     - git submodule update --init --recursive
 machine:
   node:
-    version: 0.12
+    version: 6.9.0
   pre:
     - git config --global user.name "circleci"
     - git config --global user.email "circleci@users.noreply.github.com"
--- a/circle.yml
+++ b/circle.yml
@@ -4,7 +4,7 @@ checkout:
     - git submodule update --init --recursive
 machine:
   node:
-    version: 6.9.0
+    version: 8
   pre:
     - git config --global user.name "circleci"
     - git config --global user.email "circleci@users.noreply.github.com"
--- a/site/package.json
+++ b/site/package.json
@@ -3,12 +3,12 @@
   "version": "0.0.0",
   "private": true,
   "hexo": {
-    "version": ""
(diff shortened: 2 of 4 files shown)


abernix:
@mitar Any thoughts on this?

circle.yml
```diff
@@ -4,7 +4,7 @@ checkout:
     - git submodule update --init --recursive
 machine:
   node:
-    version: 0.12
+    version: 6.9.0
```
abernix: If we want to use `6` as the baseline, this could just be `6` and it will automatically use the latest `6`.  However, since Meteor never shipped with an 8.x version I think it's safe to bump this to `8`.  I'll adjust accordingly.

site/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "hexo": {
-    "version": "3.2.2"
+    "version": ""
```
abernix: This is a `hexo` set parameter (which I believe it uses to help in automated migrations), I'm not sure setting it to blank will do anything other than resulting it getting re-set the next time `hexo` is run, but I'm willing to try it!

abernix: Follow-up, it didn't work out: https://github.com/meteor/blaze/pull/284/commits/0d1a4e3a46a941a54f4f1fdb6542cf9ef0f46181.  No worries!

abernix:
In an effort to move https://github.com/meteor/meteor/pull/9814 along, I'm going to go ahead and merge this, as I feel it's all correct.

mitar:
Thanks!

(Sorry, I am currently swamped with some other projects.)

PR #287, opened 2018-07-20
Make Template._currentTemplateInstanceFunc read-only when possible.

benjamn:
By defining `Template._currentTemplateInstanceFunc` using a getter function (without a corresponding setter function), we can make it effectively read-only, which should prevent any outside tampering.

Besides being a reasonable safety measure, this change also appears to fix a bizarre JSC (Safari) bug that can make `Template.instance()` return `null`: https://github.com/meteor/meteor/issues/9926#issuecomment-406464957

### DIFF
--- a/packages/blaze/package.js
+++ b/packages/blaze/package.js
@@ -1,7 +1,7 @@
 Package.describe({
   name: 'blaze',
   summary: "Meteor Reactive Templating library",
-  version: '2.3.2',
+  version: '2.3.3',
   git: 'https://github.com/meteor/blaze.git'
 });
 
--- a/packages/blaze/template.js
+++ b/packages/blaze/template.js
@@ -449,24 +449,67 @@ Template.prototype.helpers = function (dict) {
     this.__helpers.set(k, dict[k]);
 };
 
-// Kind of like Blaze.currentView but for the template instance.
-// This is a function, not a value -- so that not all helpers
-// are implicitly dependent on the current template instance's `data` property,
-// which would make them dependenct on the data context of the template
-// inclusion.
-Template._currentTemplateInstanceFunc = null;
-
-Template._withTemplateInstanceFunc = function (templateInstanceFunc, func) {
-  if (typeof func !== 'function')
-    throw new Error("Expected function, got: " + func);
-  var oldTmplInstanceFunc = Template._currentTemplateInstanceFunc;
-  try {
-    Template._currentTemplateInstanceFunc = templateInstanceFunc;
-    return func();
-  } finally {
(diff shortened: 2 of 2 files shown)


mitar:
Hm, if you are using `defineProperty`, then you could just make the property read-only through that, no need for getter, no? Set `writable` to false?

benjamn:
Sure, but I'm also partly trying to force reads of `Template._currentTemplateInstanceFunc` to go through a getter function, since the simple property lookup in `Template.instance` seems to be unreliable in JSC/Safari (see the issue I mentioned). I wish I understood the actual reasons for that bug better. It's pretty bizarre!

mitar:
So I tried to understand that issue but I gave up. :-)

Are you sure this getter stuff will not slow things down? Because this function is like called all the time. Maybe it would be better to do this only on Safari? (Or have a reliable test to detect when it has to be done.) Instead of making it based on support for getters?

benjamn:
The other option would be to use `Object.getOwnPropertyDescriptor(Template, "_currentTemplateInstanceFunc").value` in `Template.instance` rather than relying on `Template._currentTemplateInstanceFunc`, perhaps only in Safari.

mitar:
So, I am not sure. I would delegate to you to make a decision here. I am just bringing some points here which I imagine when looking at this change. But I do not have too much time to look into this deeper.

arggh:
@mitar 's [web-rendering-framework-benchmark](https://github.com/mitar/meteor-web-rendering-framework-benchmark) could come in handy here?

arggh:
I did a quick comparison using @mitar 's benchmarking tool, seems like there is no meaningful impact on performance in either direction. (disclaimer: I used the "skip loops for faster benchmarking" -option to get some quick insight)

blaze_compare

arggh:
Here's the results from the full benchmark. The difference is negligible, mostly in favor of the newer `2.3.3` version in this PR.

blaze_full_results

benjamn:
Thanks for the performance check, @arggh, and sorry for the wait!

I would like to get this change out sooner rather than later, to fix the Safari bug. We can always revisit it later if there are negative performance consequences in browsers other than Chrome (which seems to have no performance penalty, per @arggh's measurements) and Safari (which suffers from https://github.com/meteor/meteor/issues/9926). I would be open to restricting the getter change to Safari, if client-side `navigator.userAgent` sniffing wasn't so tricky.

arggh:
I'm very grateful for this work, thank you @benjamn ! I'll run the benchmarks on all notable browsers before putting this into production and will raise an issue if there's any significant changes in performance charasteristics.

Issue #220, opened 2016-12-15
Comments inside a head tag are not removed


hwillson:
Migrated over from meteor/meteor#3677.

----

(originally reported by @mitar)

```
<head>
  {{! This should be removed }}
</head>
```

In the output static HTML file from Meteor, this tag is still present. But it should be removed.

filipenevola:
I'm closing this issue because it's too old.

If you think this issue is still relevant please open a new one.

Why? We need to pay attention to all the open items so we should keep opened only items where people are involved.

PR #289, opened 2018-08-11
Remove comments inside head tag

jankapunkt:
Referring to #220 

As far as I can see there is no support yet for any block tag code inside head, which is why I just stripped all content that begins and ends with double curly brackets.

I hope this approach is not totally wrong  :see_no_evil:

Tests ran fine locally, although I did not add any further tests that explicitly aims at block tags. If one or two units of tests are desired I may need some further help on where exactly place the tests ( I did not find test files that correspond to the files I changed ).

### DIFF
--- a/packages/templating-tools/html-scanner-tests.js
+++ b/packages/templating-tools/html-scanner-tests.js
@@ -88,13 +88,19 @@ Tinytest.add("templating-tools - html scanner", function (test) {
 
   // head with single line comments
   checkResults(
-    scanForTest("<head>\n{{! this is the title}}\n<title>Hello Comment</title>\n</head>\n\n<body>World</body>\n\n"),
+    scanForTest("<head>{{! this is the title}}\n<title>Hello Comment</title>\n</head>\n\n<body>World</body>\n\n"),
     simpleBody('"World"'),
     "<title>Hello Comment</title>");
 
   // head with multi line comments
   checkResults(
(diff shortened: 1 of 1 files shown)


jankapunkt:
Already found myself some cases where this will not work. Should I close and reopen when fixed or add commits to this PR?

mitar:
Just add commits.

jankapunkt:
First I thought this can be done using a RegEex but it tuned out that it is (as with HTML) not sufficient to use.

Now the approach is like this

1. Check if the head content contains any occurrence of brackets. If there is no bracket, return the content as is (this will be in most of the cases)
2. If there are any brackets that indicate block content / comments, parse the head contents using `SpacebarsCompiler.parse`
3. Filter the parsed elements list by the following rules
   - remove elements of type `'string'` that do not contain at least one non-whitespace character (removes all the elements that are just newlines but keeps child contents like the title content in `<title>`
   - remove elements that have a `type` property. This removes all template includes
   - filter recursively so that comments in children are also stripped
4. Call `SpacebaesCompiler.codeGen` with a new flag called `isHead` that causes the method to call `HTML.ToHTMLVisitor().visit(tree)` and generate back the proper HTML from the parsed elements
5. Assign the generated HTML to the result.

I added tests for single line comments and multi line comments. Regarding other block content I am not sure, whether an Error should be thrown or if they should be stripped, too. Currently they are stripped as well.

PR #292, opened 2019-03-20
Add back raw/endraw to tags {{ name }} and {{ score }} in intro docs

abrahamks:
The tags `{{name}}` and `{{score}}` are not properly rendered on http://blazejs.org/ homepage.
I think this is related to a previous PR #135


filipenevola:
Hi, as the CLA is not signed we can't proceed with this PR. If you want to proceed please comment here.

PR #318, opened 2021-03-15
Spacebars compiler optimisations

nathan-muir:
Hi, (@filipenevola  Reopening #279)

Found a few instances where the spacebars compiler aborts optimisations unnecessarily.

Case 1: Inside Block tags, eg `{{#if }} ... {{else}} ... {{/if}}`
Case 2: More than one level of nested elements.  (Compacting `HTML.Raw` fails)
Case 3: Inside `<tables>` (caused by a pessimistic workaround to auto-injection of `<tbody>` when using `innerHTML`.)

The resolution for Case 1 turned out messier than I had hoped. Due to block tags like `{{#markdown}}` that don't contain HTML. Rather than exposing the `textMode` prop, maybe something like `isHTML`?

----

My motivation for digging through this code, is to implement a `<template whitespace="strip">` mode [1]. Which would work similar to react (Strip any whitespace that contains a newline).

Early comparisons show that this saves around ~100kb (uncompressed) from my client bundle.

[1] https://github.com/3stack-software/blaze/commit/8562d2d7e8be9762580dc93d93c7d8bc0858a202


nathan-muir:
Ah, it pulled in a bad commit. one moment.

PR #329, opened 2021-04-26
Remove underscore from static-html

StorytellerCZ:
Remove underscore from `static-html`, there is no need to get the entire package for just one function that can be easily replaced: https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_isempty


jankapunkt (approved):
Looks good to me

filipenevola (approved):

PR #330, opened 2021-04-26
Remove some old compat and update dependencies

StorytellerCZ:
While looking for old compatibility pre Meteor 1.0 I have found few here in Blaze and have taken the liberty to remove them. At the same time I have also updated the dependencies for the affected packages. Version bump is debatable.


jankapunkt:
Looks good to me

jankapunkt (approved):
Looks good to me

packages/spacebars/package.js
```diff
@@ -1,7 +1,7 @@
 Package.describe({
   name: 'spacebars',
   summary: "Handlebars-like template language for Meteor",
-  version: '1.1.0',
+  version: '1.2.0',
```
filipenevola, line 4: Why this should be a minor if the others are patch?

filipenevola (changes requested):
As we are removing code that maybe someone is still using I believe a major would be better but at least a minor would be required for all affected pages, no?

packages/spacebars/package.js
StorytellerCZ: Fixing

StorytellerCZ:
@filipenevola That is what needs to be discussed. These are all compatibility workarounds for before 1.0 version. I'm fine with both approaches.

jankapunkt:
I would favor a major version bump if old compat is removed, too. However, I am not sure how big the implications are for packages depending on `templating@1.x` or `blaze@2.x`. Especially those "unmaintained" ones will definitely become incompatible of those deps are declared, right?

filipenevola:
> I would favor a major version bump if old compat is removed, too. However, I am not sure how big the implications are for packages depending on `templating@1.x` or `blaze@2.x`. Especially those "unmaintained" ones will definitely become incompatible of those deps are declared, right?

This is my concern as well but at least a minor bump we should do on all of them.

filipenevola (changes requested):
I still think we should have a minor bump. Why don't you agree?

distalx:
Minor version bump makes sense to me to keep away from accidentally breaking something.

StorytellerCZ:
Then I think we have a consensus for minor version bump due to older packages per @jankapunkt comment.
Or we could make this a basis for the next major version of Blaze and get in there also the modernization work that @harryadel is working on.

PR #331, opened 2021-04-27
Remove underscore

harryadel:
Please bear with me as this my first time contributing to the Blaze project.

The goal of this pull request is to:
- Remove underscore and all of its methods in the code
- ~Modernize the code base~

Both `test:ci` and `test:watch` yield no errors.

That's for the code part but the site Hexo server isn't working. I noticed it reads jsdoc comments from several files like `builtins.js`, `preamble.js` and couple more so I reverted the linting on those files but then a weird error showed up in the browser 
```
Uncaught SyntaxError: Invalid regular expression: /(?:[ Ã‚â€º][[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-PRZcf-nqry=><])|[
``` 

I tried reverting the linting on the whole project and even updating the dependencies to the latest in attempt at least to get a more readable error but in vain. 

Screenshot_2021-04-27_19-36-07

I attempted to remove the underscore dependency from site directory but then reverted my changes so you guys can have easier time debugging things.


harryadel:
So split the code modernization into its own PR?

harryadel:
@StorytellerCZ Done!

harryadel:
I'll open up the modernization PR once this one gets merged. :+1:

StorytellerCZ (approved):
Looks good to me. I'm thinking that if we can combine this together with removal of jQuery and general modernization that you proposed, then we can do major version bump.

harryadel:
>Looks good to me. I'm thinking that if we can combine this together with removal of jQuery and general modernization that you proposed, then we can do major version bump.

The removal of jQuery would definitely require a major version bump because as @jankapunkt pointed out people's Blaze code is making heavy use of internal jQuery method.
>. This requires very precise planning to ensure the API is not breaking (think of `templateInstance.$(...)`)

filipenevola (changes requested):
Hi @harryadel , thank you for the work here.

For the review to be effective we need to have PRs that are doing only one thing.

Could you keep only the underscore changes in this PR so we can move forward with it?

harryadel:
@filipenevola There's nothing but underscore changes. I've removed the modernization commits as Jan recommended.

StorytellerCZ:
@harryadel I think the issue there is that your IDE or something automatic also "fixed" lot of the code around.

packages/templating-runtime/templating.js
```diff
@@ -85,13 +87,13 @@ Template._applyHmrChanges = function (templateName) {
...
     for (var i = 0; i < Template.__pendingReplacement.length; i++) {
-      delete Template[Template.__pendingReplacement[i]]
+      delete Template[Template.__pendingReplacement[i]];
     }
 
     Template.__pendingReplacement = [];
 
     var views = Blaze.__rootViews.slice();
-    for(var i = 0; i < views.length; i++) {
+    for (var i = 0; i < views.length; i++) {
```
StorytellerCZ, line 96: This could probably be a `forEach`

StorytellerCZ (changes requested):

jankapunkt:
As far as I saw, that was mostly lint fixes which is why I thought them to be ok

harryadel:
I see so the problem lies with the extra linting. Ok, will remove it and notify you guys.

harryadel:
@filipenevola I guess it's ready now

packages/templating-runtime/templating.js
harryadel: Tried to change it but resulted in an error.

StorytellerCZ: OK, for now. We should take a look at this in detail later then.

StorytellerCZ (approved):

packages/blaze/util.js
```diff
@@ -0,0 +1,25 @@
+has = function (obj, key) {
```
filipenevola, line 1: we shouldn't expose anything globally, use exports / imports instead.

maybe it would be better to use specific modules of lodash instead of reimplementing these functions here.

harryadel:
@filipenevola friendly ping, please let me know what you think.

StorytellerCZ (approved):

packages/blaze/template.js
```diff
@@ -426,9 +431,8 @@ Blaze.TemplateInstance.prototype.subscribe = function (/* arguments */) {
  */
 Blaze.TemplateInstance.prototype.subscriptionsReady = function () {
   this._allSubsReadyDep.depend();
-
-  this._allSubsReady = _.all(this._subscriptionHandles, function (handle) {
-    return handle.ready();
+  this._allSubsReady = Object.keys(this._subscriptionHandles).every((handle) => {
```
filipenevola, line 434: It seems Object.values would be better here as you don't need the keys.

packages/blaze/view.js
```diff
@@ -874,10 +874,11 @@ Blaze._addEventMap = function (view, eventMap, thisInHandler) {
     throw new Error("View must have a DOMRange");
 
   view._domrange.onAttached(function attached_eventMaps(range, element) {
-    _.each(eventMap, function (handler, spec) {
+    Object.keys(eventMap).forEach(function (spec) {
```
filipenevola, line 877: Object.values would be better here as you don't use the key

packages/caching-html-compiler/caching-html-compiler.js
```diff
@@ -1,5 +1,7 @@
 const path = Plugin.path;
 
+const isEmpty = obj => [Object, Array].includes((obj || {}).constructor) && !Object.entries((obj || {})).length;
```
filipenevola, line 3: Why not from lodash?

packages/spacebars-compiler/codegen.js
```diff
@@ -56,7 +56,7 @@ var additionalReservedNames = ["body", "toString", "instance",  "constructor",
 export function isReservedName(name) {
   return builtInBlockHelpers.hasOwnProperty(name) ||
     builtInTemplateMacros.hasOwnProperty(name) ||
-    _.indexOf(additionalReservedNames, name) > -1;
+    additionalReservedNames.indexOf(name) > -1;
```
filipenevola, line 59: Why not includes?

packages/templating-runtime/dynamic.js
```diff
@@ -1,3 +1,14 @@
+has = function (obj, key) {
```
filipenevola, line 1: Why not from lodash?

packages/templating-tools/compile-tags-with-spacebars.js
```diff
@@ -12,6 +12,8 @@ export function compileTagsWithSpacebars(tags, hmrAvailable) {
   return handler.getResults();
 }
 
+const isEmpty = obj => [Object, Array].includes((obj || {}).constructor) && !Object.entries((obj || {})).length;
```
filipenevola, line 15: Why not from lodash?

filipenevola (changes requested):
There are still places where you should use a different Object utility and also some places where you should use lodash.

harryadel:
So the problem that was causing the site to fail was the non existence of files within the themes directory which is fixed by `git submodule update --init --recursive`. Should've cloned recursively as suggested in the [`constributing.md`](https://github.com/meteor/blaze/blob/master/CONTRIBUTING.md#how-to-run-your-app-with-the-future-version-of-blaze-to-test-it). :sweat_smile:   

There's a tiny problem with installing modules for the site though which I'll fix in a different PR.

harryadel:
@filipenevola @StorytellerCZ

StorytellerCZ (approved):

filipenevola (approved):
❤️

PR #334, opened 2021-05-08
Faster fragment parsing

znewsham:
Happy to see Blaze getting some Love - I've been using this change in production for a good long while, figured it was maybe time to merge.

Whenever you render raw HTML in `{{{}}}` and whenever you render any static HTML that has no `{{}}` in it (e.g., `<li>Test</li>`) ` jquery's `parseHTML` is called, which generates a new document context. This causes some pieces of jquery to try and figure out the compatibility (IE/ Safari quirks, etc) of the current document EVERY time an element like this is rendered. 

I'm 99% sure we could just return `document` from `DOMBackend.getContext` - but this is "more correct". The code is taken from `jq.parseHTML`

I'm wondering if anyone can see any issues with this? Possibly if rendering into an iframe?

### DIFF
--- a/packages/blaze/dombackend.js
+++ b/packages/blaze/dombackend.js
@@ -11,7 +11,6 @@ DOMBackend._$jq = $jq;
 
 
 DOMBackend.getContext = function() {
-  return document;
   if (DOMBackend._context) {
     return DOMBackend._context;
   }
--- a/packages/blaze/dombackend.js
+++ b/packages/blaze/dombackend.js
@@ -9,13 +9,33 @@ if (! $jq)
 
 DOMBackend._$jq = $jq;
 
+
+DOMBackend.getContext = function() {
+  return document;
+  if (DOMBackend._context) {
+    return DOMBackend._context;
+  }
+  if ( DOMBackend._$jq.support.createHTMLDocument ) {
+    DOMBackend._context = document.implementation.createHTMLDocument( "" );
+
+    // Set the base href for the created document
+    // so any parsed elements with URLs
+    // are based on the document's URL (gh-2965)
+    const base = DOMBackend._context.createElement( "base" );
+    base.href = document.location.href;
+    DOMBackend._context.head.appendChild( base );
+  } else {
+    DOMBackend._context = document;
+  }
+  return DOMBackend._context;
+}
 DOMBackend.parseHTML = function (html) {
   // Return an array of nodes.
   //
   // jQuery does fancy stuff like creating an appropriate
   // container element and setting innerHTML on it, as well
   // as working around various IE quirks.
-  return $jq.parseHTML(html) || [];
+  return $jq.parseHTML(html, DOMBackend.getContext()) || [];
 };
 
 DOMBackend.Events = {


StorytellerCZ (approved):

StorytellerCZ:
I have nothing to add from myself (which counts little), but it would be nice if you could add entry to history that lets people know about the significance of the change.

znewsham:
@StorytellerCZ Happy to do so - what should the format be, should I add a new heading `v.2.4.1, 2021-May-12` and list it there?

StorytellerCZ:
@znewsham Yes, follow the format of the previous releases.

znewsham:
@StorytellerCZ done

filipenevola:
Hi @znewsham thanks for this.

Could you add some tests for these changes?

jankapunkt:
Just a comment in perspective - it has been requested some times to remove jQuery. The `parseHTML` is currently a crucial part of Blaze but relies on jQuery. Nothing wrong with this PR, I actually am grateful for this one :heart:

StorytellerCZ:
Agreed with @jankapunkt if we could do this without jQuery that would be awesome!

jankapunkt:
@znewsham is there anything you need in order to improve on this? If you add me to your repo I can help to continue this PR

znewsham:
@jankapunkt Unfortunately I'm completely swamped at the moment, and removing jQuery has no benefits to me since I use jQuery anyway :D so this isn't a high priority for me. 

Thanks for offering to help! I added you to the repo!

jankapunkt:
I will not remove jQuery yet but I can fix the merge conflicts and add some tests so the pr can be merged soon 🙂

StorytellerCZ:
@jankapunkt please proceed.

packages/blaze/dombackend.js
```diff
@@ -9,13 +9,32 @@ if (! $jq)
 
 DOMBackend._$jq = $jq;
 
+
+DOMBackend.getContext = function() {
```
jankapunkt, line 13: Please add documentation to this function. I know many internal functions are not documented but we need better documentation of internals to have more people to easily pick up the development.

packages/blaze/dombackend.js
```diff
@@ -9,13 +9,32 @@ if (! $jq)
...
+    // Set the base href for the created document
+    // so any parsed elements with URLs
+    // are based on the document's URL (gh-2965)
+    const base = DOMBackend._context.createElement( "base" );
+    base.href = document.location.href;
+    DOMBackend._context.head.appendChild( base );
+  } else {
+    DOMBackend._context = document;
+  }
+  return DOMBackend._context;
```
jankapunkt, line 29: If `DOMBackend._context` exists it is already returned at line 14, therefore this line would return `undefined` right? In such case it should return `undefined` so it's immediately clear that nothing is returned.

jankapunkt (changes requested):

packages/blaze/dombackend.js
znewsham: It gets set on lines 25 or 27. I think I just extracted this whole thing from jQuery

jankapunkt:
Hey @znewsham I am getting back at this one, finally. What would I need to do in order to tests this one in a running app using Blaze?

znewsham:
@jankapunkt assuming you've cloned the package into a local app's `packages` directory for testing, rendering raw HTML from a helper:

```
//js
Template.whatever.helpers({rawHtml() { return "<span>Hello</span>";}})
//html
<template name="whatever">
{{{rawHtml}}}
<span>Hello</span>
</template>
```

should trigger the new behaviour (e.g., any element where you don't use `{{}}` in the tag)

jankapunkt (approved):

Issue #339, opened 2021-06-07
Accept iterable in each block


jankapunkt:
Currently I can only pass arrays, cursors or falsey values to `#each`, otherwise I get this error:

```
 Error: {{#each}} currently only accepts arrays, cursors or falsey values.
```

However, there are widely accepted standard implementations like `Set` or `Map` that implement `iterable` and can easily be iterated using either their builtin `forEach` or using `for .. of`.

This would ease up many use-cases, where I currently have to switch between Map/Set and Array.

Involved tasks:

- Research where / how `#each` iterates and if `iterable` can be used
- Implement an iteration using `iterable`

antoinep92:
Do you expect {{#each "string"}} to iterate over each character then ? Or should we only whitelist some classes like Map and Set ?

antoinep92:
I can make a PR for this, but we need to decide the behavior. I see 3 options:
- accept any iterable
- accept any iterable except blacklisted types (string, etc.)
- accept anything all whitelisted types (but add Map and Set to the current whitelist)

jankapunkt:
Great @antoinep92 !

I would rather be open here, since iterable is a well-proven interface in web-dev and I assume someone implementing it, knows what they are doing.

However, I can see, that some things just don't make sense or could cause confusion (beginners accidentally iterate over string and wonder why the see a list of characters).

I would therefore go for option two, accept any + blacklist primitives (string, number, bigint, boolean, undefined, symbol, and null), function and, of course all Objects/instances that do not implement iterable.

This is of course just my current opinion but I think it might be beneficial to let users implement their own iterable structures if they want :-D

antoinep92:
I also think it's the better option.

Since number, undefined, etc. do not implement iterable there is no need to explicitly blacklist them. From what I've seen of the code, it seems simpler and safer to use Array.from, this incurs a minor performance penalty since it will make a shallow copy of the iterable, but for this use-case (DOM-rendering), I think this is negligible and an acceptable tradeoff.

I'll try to draft a PR this week.

jankapunkt:
I could provide some Tests Code for some iterables

- Map
- Set
- linked list
- binary tree
- es6 Generator function

antoinep92:
@jankapunkt I just created a PR for this. I didn't add new tests, so feel free !
Also if you can have a look at my diff and see if you spot anything, that would be appreciated.
[...]

PR #341, opened 2021-06-15
add support for arbitrary iterables in #each templates

antoinep92:
Resolves meteor/blaze#339
Moved from https://github.com/meteor/meteor/pull/11477 since `observe_sequence` was moved.

As discussed in the issue, this extends the range of supported {{#each}} arguments to any object implementing the iterable protocol except strings. To minimize changes and risks, these arguments are converted to Arrays, so the iteration mechanics remains unchanged.
IE is not supported.


antoinep92:
Note: I managed to pass all tests locally by running `meteor update` which bumped versions in `test-app/.meteor/` (mongo version was an issue) but didn't commit those changes as this will probably have other consequences and is outside the scope of the PR.

StorytellerCZ:
@antoinep92 Thanks for the heads up. I have that fixed in my PR #330

Issue #347, opened 2021-07-02
observe-sequence@1.0.18 syntax broken on IE 11


dougfabris:
Recently we updated the meteor version e we realize the `observe-sequence@1.0.18` is broken on Internet Explorer 11

We believe is something related to the syntax, not supported, present here: https://github.com/meteor/blaze/blob/master/packages/observe-sequence/observe_sequence.js#L170

As a workaround, we are doing a downgrade to the `v1.0.16`: https://github.com/RocketChat/Rocket.Chat/pull/22557
We hope you guys can have a fix for that soon, thanks!

StorytellerCZ:
Confirmed, `for-of` loop is not compatible with IE at all. I'll get a fix out today.

StorytellerCZ:
ping @antoinep92 who implemented this originally.

antoinep92:
Sorry about that. I'll submit a fix tomorrow

antoinep92:
Hum ok I saw your fix. Indeed it should work in IE, but breaks my code for non-array iterables. I thought about using `Array.from` but that's [not supported either](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#browser_compatibility).
Is there some kind of library you use that provides a polyfill or equivalent to `Array.from`, or should I implement it "by hand" ? :thinking:

antoinep92:
After giving this some more thought: I can use `Array.from` safely, since it won't trigger a parse error in IE 11 (as `for .. in` did) and iterables aren't supported on IE either, so the branch with `Array.from` will be unreachable in IE. I'll submit a PR shortly.

PR #349, opened 2021-07-06
fix regression: non array iterables were always treatead as empty

antoinep92:
Issue #347  was fixed by @StorytellerCZ in PR #348, but the fix broke the rich error message (#each value display) for non-array iterables. This PR fixes that regression. The fix works as follow:
- we add a special case for displaying arrays, that's compatible with IE9+
- the *iterable but not array* case is unreachable under IE since, so we use Array.from there


StorytellerCZ:
Thanks a lot @antoinep92 !
I will try to release this in the next two hours.

PR #351, opened 2021-08-14
Eliminate whitespace in Template.dynamic

Torgen:
By moving the whitespace from between the handlebars to inside them, six newlines are eliminated from each instantiation of Template.dynamic.. This is important for cases where you need to have no whitespace between the dynamic template content and the surrounding tags.

### DIFF
--- a/packages/templating-runtime/dynamic.html
+++ b/packages/templating-runtime/dynamic.html
@@ -3,21 +3,21 @@ the template to render) and an optional `data` property. If the `data`
 property is not specified, then the parent data context will be used
 instead. Uses the __dynamicWithDataContext template below to actually
 render the template. -->
-<template name="__dynamic">
-  {{checkContext}}
-  {{#if dataContextPresent}}
-    {{# __dynamicWithDataContext}}{{> Template.contentBlock}}{{/__dynamicWithDataContext}}
-  {{else}}
-    {{! if there was no explicit 'data' argument, use the parent context}}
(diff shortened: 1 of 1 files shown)


StorytellerCZ:
@Torgen can you provide an example between the current approach and your fix? Wouldn't minification take care of that?

Torgen:
https://github.com/Torgen/codex-blackboard/blob/gravatar-mentions/text_chunk.html
I essentially duplicated Template.dynamic with the whitespace removed to get the behavior I want.
Think of something like Slack or Twitter, where there's a plain-text way of @-mentioning someone or adding a hashtag, but you potentially want that rendered with a reactive component; often those apis include both a plaintext version and a chunked version, or have other metadata that marks the index ranges for special handling. For example, if you @-mention someone, maybe you want to show their avatar (which they're allowed to change) and something to indicate whether they're online. If there's a punctuation mark after the @-mention, it should stay attached, but the body of Template.dynamic having whitespace it it makes the browser think you want to put space between them.
I could always use Blaze.toHTML on my chunks and insert the entire message into the page with SafeString or {{{}}}, and if any of the components of the message has a reactive update, the entire thing will get re-rendered. I'm not sure if you want to encourage that.

StorytellerCZ:
Alright, will include this in Blaze 2.6

filipenevola:
@StorytellerCZ I believe this should be documented as a possible breaking change.

PR #359, opened 2022-01-22
Remove underscore from observe sequence

harryadel:
As I'm wrapping up https://github.com/meteor/meteor/pull/11869 [Test Group 3](https://app.circleci.com/pipelines/github/meteor/meteor/2782/workflows/150e8e89-1d4d-4bbd-b448-e05355b496c4/jobs/77335) is failing because it runs non-core packages and observe-sequence [requires underscore](https://github.com/meteor/blaze/blob/master/packages/observe-sequence/package.js#L20) which is no longer around since now it's now deprecated. There's no need for publishing a new version of Blaze or Observer-sequence; this PR just needs to be merged in dev branch so in meteor repo the latest commit can pulled. Coincidentally, I'm glad the the underscore removal was done here first as it'd not have worked if it had been done in Meteor repo first. :sweat_smile:

### DIFF
--- a/packages/observe-sequence/observe_sequence.js
+++ b/packages/observe-sequence/observe_sequence.js
@@ -1,3 +1,6 @@
+import isObject from 'lodash.isobject';
+import has from 'lodash.has';
+
 const warn = function () {
   if (ObserveSequence._suppressWarnings) {
     ObserveSequence._suppressWarnings--;
@@ -15,7 +18,7 @@ const warn = function () {
 // subclassed arrays: instanceof Array === true, _.isArray(arr) === false
 // see specific tests
 function isArray(arr) {
-  return arr instanceof Array || _.isArray(arr);
+  return arr instanceof Array || Array.isArray(arr);
 }
 
 const idStringify = MongoID.idStringify;
@@ -94,7 +97,7 @@ ObserveSequence = {
         if (activeObserveHandle) {
           // If we were previously observing a cursor, replace lastSeqArray with
           // more up-to-date information.  Then stop the old observe.
-          lastSeqArray = _.map(lastSeq.fetch(), function (doc) {
+          lastSeqArray = lastSeq.fetch().map(function (doc) {
             return {_id: doc._id, item: doc};
           });
           activeObserveHandle.stop();
@@ -204,9 +207,13 @@ const badSequenceError = function (sequence) {
                    sequenceGotValue(sequence));
 };
 
+const isFunction = (func) => {
+  return typeof func === "function";
+}
+
 const isStoreCursor = function (cursor) {
-  return cursor && _.isObject(cursor) &&
-    _.isFunction(cursor.observe) && _.isFunction(cursor.fetch);
+  return cursor && isObject(cursor) &&
+    isFunction(cursor.observe) && isFunction(cursor.fetch);
 };
 
 // Calculates the differences between `lastSeqArray` and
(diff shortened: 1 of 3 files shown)


StorytellerCZ (approved):
LGTM

StorytellerCZ:
Let's merge this into 2.6 release and release 2.6. What do you think @filipenevola ?

fredmaiaarantes (approved):
LGTM

packages/observe-sequence/observe_sequence_tests.js
```diff
@@ -678,7 +678,7 @@ Tinytest.add('observe-sequence - vm generated number arrays', function (test) {
 });
 
 Tinytest.add('observe-sequence - number arrays, _id:0 correctly handled, no duplicate ids warning #4049', function (test) {
-  var seq = _.map(_.range(3), function (i) { return { _id: i}; });
+  var seq = [...Array(2).keys()].map(function (i) { return { _id: i}; });
```
renanccastro, line 681: they are not equivalent, you are missing the third element here

packages/observe-sequence/observe_sequence_tests.js
```diff
@@ -45,11 +45,11 @@ runOneObserveSequenceTestCase = function (test, sequenceFunc,
...
       firedCallbacks.splice(i, 0, obj);
     },
-    removedAt: function () {
-      firedCallbacks.push({removedAt: _.toArray(arguments)});
+    removedAt: function (...args) {
+      firedCallbacks.push({removedAt: args});
     },
     movedTo: function () {
-      firedCallbacks.push({movedTo: _.toArray(arguments)});
+      firedCallbacks.push({movedTo: args});
```
renanccastro, line 52: there is no mention to args in this function parameters

renanccastro (changes requested):

packages/observe-sequence/observe_sequence_tests.js
harryadel: Thanks for noticing these @renanccastro. You may review again now.

harryadel:
Any update on this and the 2.6 release? @StorytellerCZ @renanccastro @filipenevola

StorytellerCZ:
@fredmaiaarantes @jankapunkt let's include this in 2.6?

jankapunkt (commented):
I would generally add it to 2.6 but I'd like to discuss one more this PR (see my comment on the code).

If this is to be a showstopper (due to further discussion) then we might publish 2.6 with this PR as-is and move the final removal of lodash (see my comment on the code) to 3.0?

packages/observe-sequence/observe_sequence.js
```diff
@@ -1,3 +1,6 @@
+const isObject = Npm.require('lodash.isobject');
+const has = Npm.require('lodash.has');
+
```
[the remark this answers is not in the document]
StorytellerCZ: I'm for replacement. Less dependencies is better.

StorytellerCZ:
@harryadel can you please address @jankapunkt comment? Then we can merge this into 2.6 and release.

harryadel:
Right away!

StorytellerCZ:
@harryadel this seems to be the issue of the failing tests:
```
TypeError: posNew.forEach is not a function
    at diffArray
```

harryadel:
It's weird that some errors weren't caught by the CI before :thinking:

Issue #213, opened 2016-12-13
Blaze: Uncaught Error: Must be attached


hwillson:
Migrating this over from meteor/meteor#2981.

----

(originally reported by @mitar)

See [example project](https://github.com/mitar/meteor-issue2981). When clicking on "Click Me" button, there is an uncaught Error: `Must be attached`.

```
Uncaught Error: Must be attached domrange.js:337
DOMRange.containsElement domrange.js:337
(anonymous function) view.js:741
(anonymous function) events.js:47
jQuery.event.dispatch jquery.js:4624
elemData.handle jquery.js:4292
```

bastiW:
is there any workaround for this?

mitar:
Yes, [see discussion here](https://github.com/meteor/meteor/issues/2981#issuecomment-87654677).

aadamsx:
Ha, I just ran into this for the first time today!

rtcwp07:
I'm probably the last person still using Blaze, but for posterity I will add here that I encountered this error when I was using jQuery to get a field value for a method call, and solved it by making sure that I used the template.$ jQuery call instead.

dariovillalta:
@rtcwp07 cheers man, could you please elaborate what you mean with "the template.$ jQuery call"? Thanks!

dariovillalta:
@dr-dimitru nice explaining, thank you so much!

snajjar:
No update after 2 years? This keeps happening a lot (I use external libraries that are based on jquery)...

filipenevola:
I'm closing this issue because it's too old.

If you think this issue is still relevant please open a new one.

Why? We need to pay attention to all the open items so we should keep opened only items where people are involved.

lynchem:
I think this is worth reopening @filipenevola  as it's still happening a LOT and it would be a shame to lose all the history. It's already been evicted once. 

I don't see anything relating to jquery in the callstack for our project. The two most common call stacks I see are as follows: 


> Error: Must be attached
>     at meteor://💻app/packages/blaze/domrange.js:191:10
>     at firstNode (meteor://💻app/packages/blaze/template.js:171:38)
>     at _currentTemplateInstanceFunc (meteor://💻app/packages/blaze/template.js:559:16)
>     our code
>     at apply (meteor://💻app/imports/startup/client/blaze-exceptions.js:13:21)
>     at apply (meteor://💻app/packages/blaze/exceptions.js:51:15)
[...]

PR #366, opened 2022-04-14
Prevent firing callbacks until members are destroyed

jankapunkt:
I found a few errors related to #213 (cc @lynchem) when I moved to a new route and a new template is drawn.

I debugged my Templates and found that child templates of the page I "left" were not destroyed, while the page main template was destroyed correctly and the child template was also in `_domrange.members`.

So I started a deeper debugging session and found that `Blaze._fireCallbacks(view, 'destroyed')` was placed before `view._domrange.destroyMembers` and prevented the members from being removed, because the whole method just aborted after `_fireCallbacks` was called.

Placing it after them removes the members and the errors are gone, while at least our app tests are running fine.

I can't say if this is related to the errors in #213 but I can say that this was definitely a memory leak.
I also don't know where I should place tests for this to prevent future regressions (especially if we do a lot of rewrites for 3.0). Please help whoever can.

### DIFF
--- a/packages/blaze/view.js
+++ b/packages/blaze/view.js
@@ -489,14 +489,19 @@ Blaze._destroyView = function (view, _skipNodes) {
     return;
   view.isDestroyed = true;
 
-  Blaze._fireCallbacks(view, 'destroyed');
 
   // Destroy views and elements recursively.  If _skipNodes,
   // only recurse up to views, not elements, for the case where
   // the backend (jQuery) is recursing over the elements already.
 
-  if (view._domrange)
-    view._domrange.destroyMembers(_skipNodes);
+  if (view._domrange) view._domrange.destroyMembers(_skipNodes);
+
+  // XXX: fire callbacks after potential members are destroyed
+  // otherwise it's tracker.flush will cause the above line will
+  // not be called and their views won't be destroyed
+  // Involved issues: DOMRange "Must be attached" error, mem leak
+  
+  Blaze._fireCallbacks(view, 'destroyed');
 };
 
 Blaze._destroyNode = function (node) {


StorytellerCZ (approved):
LGTM

Blaze 2.6.1?

jankapunkt:
I don't know, I tested for our app but I have no comparison or other setups that might break. Maybe we get some more people to test, especially those, having trouble with #213

StorytellerCZ:
Let's give them some time then. I would like to see it in 2.6.1 if there are no complains.

StorytellerCZ:
@jankapunkt what should be the line for changelog?

jankapunkt:
@StorytellerCZ 

- fix destroying child templates when parent is destroyed; prevents memory leak and DOMRange "not attached" error

lynchem:
That's great @jankapunkt , good detective work. We've never found a way of reproducing #213 so can't really help with tests. We can give it a quick test in staging and if all is order launch it. We'll know after a couple of days if it's worked as it fires pretty much daily.

And the removal of a core memory leak like this is also great news 👏🏻 
When do you expect to launch it?

jankapunkt:
From my pov we could do an RC release asap which is great for testing but that's not my call. @fredmaiaarantes what do you think?

fredmaiaarantes:
> From my pov we could do an RC release asap which is great for testing but that's not my call. @fredmaiaarantes what do you think?

I also think it's worth it. 

cc @denihs

denihs (approved):

denihs:
I'm going to merge this to the branch release-2.6.1 to centralize everything there.

denihs:
Guys, blaze@2.6.1-rc.0 is ready to be tested. https://github.com/meteor/blaze/pull/371

jankapunkt:
Awesome! cc @lynchem I hope it will fix your issue as well. For the memory leak I will create a small test project for profiling next week.

pengsiya:
This change has caused a regression on select2.

Template.selection.destroyed = function () {
    var select2 = $('#selection-id');
    if (select2.hasClass('select2-hidden-accessible')) {
        select2.select2('destroy');
    }
};

Before blaze 2.6.1, this works and it successfully removed the select2 input field.

After blaze 2.6.1, blaze tries to destroy the select2, however, it does not fully kill the span input field attached to it, yet the "select2" data attribute no longer bind to the select2 dom element, therefore when trying to kill the input field by calling .select2('destroy') leads to the following error:

  The select2('destroy') method was called on an element that is not using Select2.

The issue here is that the destroy implemented in blaze partially destroyed "select2" and we will have to put in workaround to achieve the original behavior.

PR #370, opened 2022-04-14
Remove tests for removed API in spacebars, fix #369

### LINKED ISSUE #369: spacebars-tests packages still uses removed code
https://github.com/meteor/blaze/blob/da072c5a01ac84d107df47aa86997bf41e8a06ea/packages/spacebars-tests/old_templates.js#L15

StorytellerCZ:
As describe in #369 old tests for old templates API that has been remove in v2.6 have not been removed. As far as I have searched this should be the only thing that needs to be removed.


StorytellerCZ:
Together with #366 it could be made part of Blaze 2.6.1. Thoughts?

jankapunkt:
Just a question for understanding, shouldn't they have thrown errors after backwards compatibility has been removed? Is there maybe still old code?

harryadel:
Why remove files entirely? Why not keep the tests but change the APIs used? Also `Template.__define__` got removed entirely!! This is a breaking change for no reason. It's ok to remove `UI.body` and the likes since they've alternatives but not the rest!

StorytellerCZ:
This is the code for `Template.__define__`:
```js
// XXX COMPAT WITH 0.8.3
Template.__define__ = function (name, renderFunc) {
  Template.__checkName(name);
  Template[name] = new Template("Template." + name, renderFunc);
  // Exempt packages built pre-0.9.0 from warnings about using old
  // helper syntax, because we can.  It's not very useful to get a
  // warning about someone else's code (like a package on Atmosphere),
  // and this should at least put a bit of a dent in number of warnings
  // that come from packages that haven't been updated lately.
  Template[name]._NOWARN_OLDSTYLE_HELPERS = true;
};
```
I'm fine with putting it back, but I would probably make a change and show some sort of a warning at the very least.

harryadel:
Great work :clap:

StorytellerCZ:
@fredmaiaarantes @denihs this is also ready to be included in Blaze 2.6.1

denihs:
Hey @StorytellerCZ, published in the new version 2.6.1-rc.1.

PR #376, opened 2022-06-19
Modify 'Must be attached' error to be more descriptive

harryadel:
This should fix https://github.com/meteor/blaze/issues/213

### DIFF
--- a/packages/blaze/domrange.js
+++ b/packages/blaze/domrange.js
@@ -333,7 +333,9 @@ DOMRange.prototype.destroy = function (_skipNodes) {
 };
 
 DOMRange.prototype.containsElement = function (elem, selector, event) {
-  const templateName = this.view.name?.split('.')[1];
+  const templateName = this.view?.name
+  ? this.view.name.split('.')[1]
+  : 'unknown template';
   if (! this.attached)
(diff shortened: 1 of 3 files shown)


harryadel:
cc @jankapunkt @StorytellerCZ

packages/blaze/domrange.js
```diff
@@ -332,9 +332,11 @@ DOMRange.prototype.destroy = function (_skipNodes) {
   DOMRange._destroy(this, _skipNodes);
 };
 
-DOMRange.prototype.containsElement = function (elem) {
+DOMRange.prototype.containsElement = function (elem, selector, event) {
+  const templateName = this.view.name?.split('.')[1];
```
jankapunkt, line 336: I would propose to add some fallback string in case we can't resolve the name:

```js
const templateName = this.view?.name
  ? this.view.name.split('.')[1]
  : 'unknown template'
```

jankapunkt (changes requested):

jankapunkt (approved):

denihs:
@jankapunkt should be able to do it now. But to release a new version still have to be someone from inside. I can check this later with the team on how we can grant access to this part, in case @jankapunkt also wants to take care of releases.

harryadel:
Fine by me, thank you @denihs.

PR #378, opened 2022-06-20
[spacebars-compiler] Update uglify-js to 3.16.1

harryadel:
Update [uglify](https://github.com/mishoo/UglifyJS) to https://github.com/mishoo/UglifyJS/releases/tag/v3.16.1. There're no changelogs to work with to understand the transition pre 3.x so we need to be a little bit cautious here and release a beta version for testing before the full release.


harryadel:
@jankapunkt What do you think?

jankapunkt:
I would put this to 3.0 unless there is a string reason to merge it fast (like any know vulnerabilities etc.)

harryadel:
Yeah, I absolutely agree. This PR is by no means a small change.

PR #381, opened 2022-07-24
docs (Overview) add instructions on how to install `jQuery`.

### LINKED ISSUE #360: blaze is missing `jquery` in its dependency list.
Starting a blank Meteor app with no packages, then running `meteor add accounts-ui` results in dependencies for `accounts-ui` being installed, including `blaze`. However, `blaze` code does not have `jquery` listed as dependency, but it tries to use `jQuery` at runtime.

```
Uncaught Error: jQuery not found
```

distalx:
fixes #360 .
 - add instructions on how to install `jQuery`.
 
 This is how it appears.
image

### DIFF
--- a/site/source/index.md
+++ b/site/source/index.md
@@ -11,6 +11,39 @@ Blaze is a Meteor-only package for now. Soon we will have Blaze on npm so you ca
 
 Each new Meteor project you create has Blaze included (the `blaze-html-templates` package).
 
+## Prerequisite
+
+Earlier, `jQuery` was bundled with Blaze. In later versions Blaze does not have `jQuery` as a direct dependency. With some legacy Blaze packages you might encounter the following error.
+
+```js
+Uncaught Error: jQuery not found
+```
+
+To mitigate this error it is advised to install `jQuery` either as an `npm` or as a `atmospherejs` package.
+
+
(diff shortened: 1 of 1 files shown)


jankapunkt:
Thanks a lot @distalx  I have two things to comment:

- is it mandatory to import `$` or is it possible with both methods to use `$` from the global scope? If so, it should be added as a note
- would importing both methods (atmosphere AND npm) result in two jquery installations being present in the client bundle? If so, that should be mentioned to prevent people bloating their client bundles

distalx:
@jankapunkt 

> is it mandatory to import $ or is it possible with both methods to use $ from the global scope? If so, it should be added as a note

This actually depends on the packages. In most case importing $ in `client/main.js` would solve the issue. In some case they might have to attach `$` to window object. i.e. `window.$ = $` 
I've added that line as a starting point for people who might not know where to put their global imports. But we can remove that if you think it is not appropriate or confusing.


> would importing both methods (atmosphere AND npm) result in two jquery installations being present in the client bundle? If so, that should be mentioned to prevent people bloating their client bundles

They could choose either one of the methods!!

jankapunkt (approved):

StorytellerCZ (approved):

PR #391, opened 2022-08-26
Deprecate UI package

jankapunkt:
Added the deprecated flag to the package.js file

Not sure why the other file from the PR #381 is added here, though :eyes:

### DIFF
--- a/packages/ui/package.js
+++ b/packages/ui/package.js
@@ -2,7 +2,8 @@ Package.describe({
   name: 'ui',
   summary: "Deprecated: Use the 'blaze' package",
   version: '1.0.13',
-  git: 'https://github.com/meteor/blaze.git'
+  git: 'https://github.com/meteor/blaze.git',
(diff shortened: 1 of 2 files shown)


StorytellerCZ (approved):

jankapunkt:
@StorytellerCZ should we target 3.0 or 2.6.2?

packages/ui/package.js
```diff
@@ -2,7 +2,8 @@ Package.describe({
   name: 'ui',
   summary: "Deprecated: Use the 'blaze' package",
   version: '1.0.13',
```
StorytellerCZ, line 4: We should probably bump the version here and publish the package so that the warning gets out.

StorytellerCZ (changes requested):

StorytellerCZ:
Good question @jankapunkt. I think version 3 is more appropriate for this.

Issue #367, opened 2022-04-14
Move codebase to ES6


jankapunkt:
This would be a big task but I think there will be lots of benefits in terms of code-readability (especially with template-literals) and a few performance improvements either.

This could be also distributed to PR-per-package:

- [ ] blaze
- [ ] blaze-tools
- [ ] caching-html-compiler
- [ ] html-tools
- [ ] htmljs
- [ ] observe-sequence
- [ ] spacebars
- [ ] spacebars-compiler
- [ ] spacebars-tests
- [ ] templating-compiler
- [ ] templating-runtime
- [ ] templating-tools

In an optimum there would be no changes to the functionality at all.

StorytellerCZ:
We could also go one step further and move to TypeScript. :smile:

jankapunkt:
We could but I have no TS experience at all so to me it would take much more time to get into the ts thinking

StorytellerCZ:
Alright let's take it step by step.

jankapunkt:
I think a good starting point for moving towards TS would be to add type definitions at least, so intellisense can pick it up better.

rashmod:
I'm willing to do this.
I'm new to open source and this seems like a good issue to resolve. Is there anything that I must know about this project or anything else?

rashmod:
I will start working on this now. But @jankapunkt you were talking about the linter to use.....?

jankapunkt:
I just checked and the main Meteor project uses the @quave/quave config (see their package.json) which is maintained by @filipenevola so I think we should also adapt to use it.

What do you think @StorytellerCZ @denihs

denihs:
No objection from me

rashmod:
I'm having some issues while setting up the local env. I used nvm to change to node v14.20.0 and somehow installed meteor (faced many problems with this too). But when following the steps in the readme `meteor npm install` for setting local env I get the error saying package.json not found.

sidenote: my primary cli git bash does not work with meteor. "bash: meteor: command not found" even tho it works using PowerShell or cmd

jankapunkt:
I am unfortunately only on Linux and MacOS so I cannot be of much help regarding windows setup. Have you checked the Meteor forums about these issues? IIRC there should be a few solved topics on windows issues regarding installation of Meteor.

Can anyone else help out here?

rashmod:
The error I keep getting is that there is no package.json file. And according to my beginner-level understanding, you need a package.json to do npm install and stuff

[...]

PR #397, opened 2022-10-13
Refactored blaze to move ES6 version

klablink:
As requested in #367, I started to migrate the code to ES6 for the blaze package.
All tests are passing, please review and let me know for required changes.

### DIFF
--- a/packages/blaze/backcompat.js
+++ b/packages/blaze/backcompat.js
@@ -1,3 +1,6 @@
+/* global Blaze UI Handlebars ReactiveVar */
+/* eslint-disable no-global-assign */
+
 UI = Blaze;
 
 Blaze.ReactiveVar = ReactiveVar;
@@ -10,9 +13,9 @@ Handlebars._escape = Blaze._escape;
 
 // Return these from {{...}} helpers to achieve the same as returning
 // strings from {{{...}}} helpers
-Handlebars.SafeString = function(string) {
+Handlebars.SafeString = function (string) {
   this.string = string;
 };
-Handlebars.SafeString.prototype.toString = function() {
+Handlebars.SafeString.prototype.toString = function () {
   return this.string.toString();
 };
--- a/packages/blaze/dombackend.js
+++ b/packages/blaze/dombackend.js
@@ -1,11 +1,13 @@
-var DOMBackend = {};
+/* global Blaze jQuery Package */
+/* eslint-disable import/no-unresolved, no-global-assign, no-param-reassign */
+
+const DOMBackend = {};
 Blaze._DOMBackend = DOMBackend;
 
-var $jq = (typeof jQuery !== 'undefined' ? jQuery :
-           (typeof Package !== 'undefined' ?
-            Package.jquery && Package.jquery.jQuery : null));
-if (! $jq)
-  throw new Error("jQuery not found");
+const $jq = (typeof jQuery !== 'undefined' ? jQuery :
+  (typeof Package !== 'undefined' ?
+    Package.jquery && Package.jquery.jQuery : null));
+if (!$jq) throw new Error('jQuery not found');
 
 DOMBackend._$jq = $jq;
 
@@ -22,18 +24,18 @@ DOMBackend.Events = {
   // `selector` is non-null.  `type` is one type (but
   // may be in backend-specific form, e.g. have namespaces).
   // Order fired must be order bound.
-  delegateEvents: function (elem, type, selector, handler) {
+  delegateEvents(elem, type, selector, handler) {
     $jq(elem).on(type, selector, handler);
   },
 
-  undelegateEvents: function (elem, type, handler) {
+  undelegateEvents(elem, type, handler) {
     $jq(elem).off(type, '**', handler);
   },
 
-  bindEventCapturer: function (elem, type, selector, handler) {
-    var $elem = $jq(elem);
+  bindEventCapturer(elem, type, selector, handler) {
+    const $elem = $jq(elem);
 
-    var wrapper = function (event) {
+    const wrapper = function (event) {
       event = $jq.event.fix(event);
       event.currentTarget = event.target;
 
@@ -43,9 +45,8 @@ DOMBackend.Events = {
       // since jQuery can't bind capturing handlers, it's not clear
       // where we would hook in.  Internal jQuery functions like `dispatch`
       // are too high-level.
-      var $target = $jq(event.currentTarget);
-      if ($target.is($elem.find(selector)))
-        handler.call(elem, event);
+      const $target = $jq(event.currentTarget);
+      if ($target.is($elem.find(selector))) handler.call(elem, event);
     };
 
     handler._meteorui_wrapper = wrapper;
@@ -55,22 +56,21 @@ DOMBackend.Events = {
     elem.addEventListener(type, wrapper, true);
   },
 
-  unbindEventCapturer: function (elem, type, handler) {
+  unbindEventCapturer(elem, type, handler) {
     type = DOMBackend.Events.parseEventType(type);
     elem.removeEventListener(type, handler._meteorui_wrapper, true);
   },
 
-  parseEventType: function (type) {
+  parseEventType(type) {
     // strip off namespaces
-    var dotLoc = type.indexOf('.');
-    if (dotLoc >= 0)
-      return type.slice(0, dotLoc);
+    const dotLoc = type.indexOf('.');
+    if (dotLoc >= 0) return type.slice(0, dotLoc);
     return type;
-  }
+  },
 };
 
 
-///// Removal detection and interoperability.
+// /// Removal detection and interoperability.
 
 // For an explanation of this technique, see:
 // http://bugs.jquery.com/ticket/12213#comment:23 .
@@ -80,17 +80,18 @@ DOMBackend.Events = {
 // which we can detect using a custom event with a teardown
 // hook.
 
-var NOOP = function () {};
+const NOOP = function () {
+};
 
 // Circular doubly-linked list
-var TeardownCallback = function (func) {
+const TeardownCallback = function (func) {
   this.next = this;
   this.prev = this;
   this.func = func;
 };
 
 // Insert newElt before oldElt in the circular list
-TeardownCallback.prototype.linkBefore = function(oldElt) {
+TeardownCallback.prototype.linkBefore = function (oldElt) {
   this.prev = oldElt.prev;
   this.next = oldElt;
   oldElt.prev.next = this;
@@ -103,7 +104,8 @@ TeardownCallback.prototype.unlink = function () {
 };
 
 TeardownCallback.prototype.go = function () {
-  var func = this.func;
+  const { func } = this;
+  // eslint-disable-next-line no-unused-expressions
   func && func();
 };
 
@@ -116,13 +118,13 @@ DOMBackend.Teardown = {
   // one of its ancestors is removed from the DOM via the backend library.
   // The callback function is called at most once, and it receives the element
   // in question as an argument.
-  onElementTeardown: function (elem, func) {
-    var elt = new TeardownCallback(func);
+  onElementTeardown(elem, func) {
+    const elt = new TeardownCallback(func);
 
-    var propName = DOMBackend.Teardown._CB_PROP;
-    if (! elem[propName]) {
+    const propName = DOMBackend.Teardown._CB_PROP;
+    if (!elem[propName]) {
       // create an empty node that is never unlinked
-      elem[propName] = new TeardownCallback;
+      elem[propName] = new TeardownCallback();
 
       // Set up the event, only the first time.
       $jq(elem).on(DOMBackend.Teardown._JQUERY_EVENT_NAME, NOOP);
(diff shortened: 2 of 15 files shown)


StorytellerCZ:
Thanks a lot @klablink! Can you please sign the CLA, so that we can proceed on this?

klablink:
> Thanks a lot @klablink! Can you please sign the CLA, so that we can proceed on this?

Hi @StorytellerCZ, I have signed the CLA now.

packages/blaze/backcompat.js
```diff
@@ -1,3 +1,6 @@
+/* global Blaze UI Handlebars ReactiveVar */
```
jankapunkt, line 1: We can drop `UI` as it's deprecated

packages/blaze/dombackend.js
```diff
@@ -80,17 +80,18 @@ DOMBackend.Events = {
 // which we can detect using a custom event with a teardown
 // hook.
 
-var NOOP = function () {};
+const NOOP = function () {
+};
 
 // Circular doubly-linked list
-var TeardownCallback = function (func) {
+const TeardownCallback = function (func) {
```
jankapunkt, line 87: Can you transform `TeardownCallback` to an ES6 `class`?

packages/blaze/domrange.js
```diff
@@ -1,24 +1,26 @@
...
+const _emptyArray = Object.freeze ? Object.freeze([]) : [];
 
 // `[new] Blaze._DOMRange([nodeAndRangeArray])`
 //
 // A DOMRange consists of an array of consecutive nodes and DOMRanges,
 // which may be replaced at any time with a new array.  If the DOMRange
 // has been attached to the DOM at some location, then updating
 // the array will cause the DOM to be updated at that location.
+// eslint-disable-next-line consistent-return
 Blaze._DOMRange = function (nodeAndRangeArray) {
```
jankapunkt, line 14: Can you make `DOMRange` an ES6 `class`? The fmethods that are not added using `.prototype` should then be `static`

packages/blaze/domrange.js
```diff
@@ -1,24 +1,26 @@
...
 
-  var members = (nodeAndRangeArray || _emptyArray);
-  if (! (members && (typeof members.length) === 'number'))
-    throw new Error("Expected array");
+  const members = (nodeAndRangeArray || _emptyArray);
+  if (!(members && (typeof members.length) === 'number')) throw new Error('Expected array');
 
-  for (var i = 0; i < members.length; i++)
-    this._memberIn(members[i]);
+  for (let i = 0; i < members.length; i++) this._memberIn(members[i]);
```
jankapunkt, line 23: You could use `members.forEach` here instead

packages/blaze/domrange.js
```diff
@@ -88,34 +85,31 @@ DOMRange._insertNodeWithHooks = function (n, parent, next) {
   // `|| null` because IE throws an error if 'next' is undefined
```
jankapunkt, line 85: @StorytellerCZ do we still want IE support?

packages/blaze/domrange.js
```diff
@@ -128,134 +122,122 @@ DOMRange.prototype.attach = function (parentElement, nextNode, _isMove, _isRepla
...
     }
   }
 };
 
 DOMRange.prototype.addMember = function (newMember, atIndex, _isMove) {
-  var members = this.members;
-  if (! (atIndex >= 0 && atIndex <= members.length))
-    throw new Error("Bad index in range.addMember: " + atIndex);
+  const { members } = this;
+  if (!(atIndex >= 0 && atIndex <= members.length)) throw new Error(`Bad index in range.addMember: ${atIndex}`);
```
jankapunkt, line 224: please add a newline between those two

packages/blaze/events.js
```diff
@@ -178,29 +194,28 @@ EventSupport.listen = function (element, events, selector, handler, recipient, g
...
       // via stop() methods is O(N^2) in the number of handlers on
       // an element.
-      for (var i = 0; i < newHandlerRecs.length; i++) {
-        var handlerToRemove = newHandlerRecs[i];
-        var info = eventDict[handlerToRemove.type];
-        if (! info)
-          continue;
-        var handlerList = info.handlers;
-        for (var j = handlerList.length - 1; j >= 0; j--) {
+      for (let i = 0; i < newHandlerRecs.length; i++) {
```
jankapunkt, line 205: We should try to avoid `continue` where possible. A potential alternative is to use `newHandlerRecs.forEach` and use `if (!info) return;` because `return` actually only ends the function early and pops it from the call stack, `continue` however creates `goto` statements

packages/blaze/lookup.js
```diff
@@ -1,3 +1,6 @@
+/* global Blaze */
+/* eslint-disable import/no-unresolved, no-cond-assign, no-global-assign, prefer-rest-params, no-param-reassign, no-multi-assign */
```
jankapunkt, line 2: Can you please try to not just silence eslint and find better alternatives for at least `no-cond-assign`, `prefer-rest-params`, `no-param-reassign` and `no-multi-assign`?

It's okay to disable `import/no-unresolved` as they are caused by Meteor's package system and we can't work around. At this point we also can't break the hole package by removing global assignments. But the other warnings are usually solveable using ES6.

jankapunkt: Let me know if you need help on a specific situation here

packages/blaze/materializer.js
```diff
@@ -1,146 +1,77 @@
...
-    // the work.  Call each "task" function, starting with the top
-    // of the stack.
-    while (workStack.length) {
-      // Note that running task() may push new items onto workStack.
-      var task = workStack.pop();
-      task();
-    }
-  }
+/* global Blaze HTML ElementAttributesUpdater Tracker */
+/* eslint-disable import/no-unresolved, no-cond-assign, no-global-assign, no-restricted-syntax, prefer-rest-params, no-param-reassign, no-multi-assign */
```
jankapunkt, line 2: Same here, let's work them out instead of just disabling

packages/blaze/package.js
```diff
@@ -24,14 +26,14 @@ Package.onUse(function (api) {
   api.export([
     'Blaze',
     'UI',
```
jankapunkt, line 28: UI can be removed as deprecated

packages/blaze/template.js
```diff
@@ -11,52 +14,53 @@ import isEmpty from 'lodash.isempty';
 // `viewKind` is a string that looks like "Template.foo" for templates
 // defined by the compiler.
 
+const HelperMap = function () {
```
jankapunkt, line 17: Can you make `HelperMap` an ES6 `class`?

packages/blaze/template.js
```diff
@@ -11,52 +14,53 @@ import isEmpty from 'lodash.isempty';
...
+
 /**
  * @class
  * @summary Constructor for a Template, which is used to construct Views with particular name and content.
  * @locus Client
  * @param {String} [viewName] Optional.  A name for Views constructed by this Template.  See [`view.name`](#view_name).
  * @param {Function} renderFunction A function that returns [*renderable content*](#Renderable-Content).  This function is used as the `renderFunction` for Views constructed by this Template.
  */
+// eslint-disable-next-line consistent-return
 Blaze.Template = function (viewName, renderFunction) {
```
jankapunkt, line 37: Can you make `Template` an ES6 class?

packages/blaze/template.js
```diff
@@ -521,26 +529,28 @@ if (canUseGetters) {
...
-  var eventMap2 = {};
-  for (var k in eventMap) {
-    eventMap2[k] = (function (k, v) {
-      return function (event /*, ...*/) {
-        var view = this; // passed by EventAugmenter
-        var data = Blaze.getData(event.currentTarget);
+  const template = this;
+  const eventMap2 = {};
+  // eslint-disable-next-line guard-for-in,no-restricted-syntax,no-unused-vars
+  for (const k in eventMap) {
```
jankapunkt, line 538: Use `Object.getOwnPropertyNames(eventMap).forEach(...)` instead

packages/blaze/view.js
```diff
@@ -41,9 +44,10 @@
  * @param {Function} renderFunction A function that returns [*renderable content*](#Renderable-Content).  In this function, `this` is bound to the View.
  */
 Blaze.View = function (name, render) {
```
jankapunkt, line 46: Can you make `View` an ES6 `class`?

jankapunkt (changes requested):
There are some things to resolve, especially the Es5 to ES6 classes and the disabled eslint warnings. Let me know if you need help at any point.

packages/blaze/backcompat.js
klablink: With this change, many compatibility tests present in old_template_test.js will be deleted, creating possible breaking chnages. Is that what we want?

packages/blaze/lookup.js
klablink: I did a code review to that effect. We have a strong impact by disabling no-param-reassign. Many of the functions do change the value of attributes in objects. 

```
  static _destroy(m, _skipNodes) {
    if (m instanceof DOMRange) {
      if (m.view) Blaze._destroyView(m.view, _skipNodes);
    } else if ((!_skipNodes) && m.nodeType === 1) {
      // DOM Element
      if (m.$blaze_range) {
        Blaze._destroyNode(m);
        m.$blaze_range = null;
      }
    }
  }

```

```
Blaze._createView = function (view, parentView, forExpansion) {
  if (view.isCreated) throw new Error("Can't render the same View twice");

  view.parentView = (parentView || null);
  view.isCreated = true;
  if (forExpansion) view._isCreatedForExpansion = true;

  Blaze._fireCallbacks(view, 'created');
};

```

In some cases we could use the argument default but it does not cover some situations now supported: null, zero. 

```
Blaze._parentData = function (height, _functionWrapped) {
  // If height is null or undefined, we default to 1, the first parent.
  if (height == null) {
    // eslint-disable-next-line no-param-reassign
    height = 1;
  }
```

We should create many local variables to eliminate the warning.

packages/blaze/backcompat.js
jankapunkt: Yes, for 3.0 we will likely introduce breaking changes. Note that it should only be breaking for the romeval of `UI`

packages/blaze/package.js
klablink: As the previous comment

packages/blaze/backcompat.js
klablink: Thank you, I remove it

packages/blaze/lookup.js
jankapunkt: We can use nullish coalescing in such situations 

```js
const parent = height ?? 1
```

and if zero is not a valid value then simply or would be sufficient, too

```js
const parent = height || 1
```

I know this may be a bit of a chore but in the long run we will be better with local variables than param reassignment

jankapunkt: For those cases where object props are updated we should use/add an update fn on the class

```js
m.updateParentView(parentView || null)
```

packages/blaze/package.js
klablink: In the meantime, I did the push with the changes without no-param-reassign. If you take a look I will also fix any other points.

packages/blaze/lookup.js
klablink: I used no-param-reassign in a few cases and for line. Whenever possible I created a class method to modify attributes. For the remaining cases I used local variables. Let me know if this is fine or if you have any other ideas.

klablink: I did the push with the changes. 

PS: I have been using Blaze happily since 2015 and have over 500 templates to maintain and bring up to version 3.0 (Meteor/Blaze), I am very concerned about the life of the project :-)

packages/blaze/domrange.js
klablink: I am waiting for the decision before starting implementation.

klablink:
I have completed the migration of all packages, tell me if there is anything to change.

klablink:
@StorytellerCZ @jankapunkt  The removal of the UI package makes the iron-router package incompatible. I cannot find it in the community-managed packages, so in my opinion it should be added. What do you think?

jankapunkt:
> The removal of the UI package makes the iron-router package incompatible
This is a sign that iron router has to be updated and that there is lots of legacy code in it and maybe even a few parts that could be vulnerable (afaik iron router can be used on the server, too, right!?)

klablink:
> > The removal of the UI package makes the iron-router package incompatible
> 
> This is a sign that iron router has to be updated and that there is lots of legacy code in it and maybe even a few parts that could be vulnerable (afaik iron router can be used on the server, too, right!?)

Right, I think many projects started with Meteor in version 0.x still use this package.  We use it more client-side, server-side we use express.

StorytellerCZ:
@klablink can you please split this PR into smaller ones by packages, so that we can review this more efficiently. Thank you!

klablink:
> @klablink can you please split this PR into smaller ones by packages, so that we can review this more efficiently. Thank you!

Hi @StorytellerCZ I can try but it is not easy because some changes are across packages.

klablink:
@StorytellerCZ I checked the possibility of splitting PR. Unfortunately, the packages are strongly interconnected, the modification of a construct implies the revision of all references. For example, the transformation of Blaze objects into ES6 classes involves the use of the new clause in the linked sources. Now, how can we proceed to simplify the verification task?

StorytellerCZ:
I think we need to split lint fixes from this PR. There area lot of files where the only change is that it has been linted. That should go into a separate PR, we merge that and that should reduce the number of files that need to be reviewed.

jankapunkt:
Yes @StorytellerCZ good point. The biggest changes, apart from linting, are:

- classes
- arrow functions
- template literals

the rest is mostly linting.

StorytellerCZ:
Lets split this us, and get this in asap.

PR #408, opened 2023-03-27
EXPERIMENTAL / CONCEPT: Sketching a pseudo-Async-supporting Blaze.

DanielDornhardt:
READER BEWARE:

This isn't really functional / final but merely a sketch of something to be viewed/discussed.

🔥🔥 THIS ISN'T TESTED, 95 % INCOMPLETE & PROBABLY NOT A GREAT IDEA!!! 🔥🔥

But I tried to write nice comments at least! 😄

### DIFF
--- a/packages/blaze/view.js
+++ b/packages/blaze/view.js
@@ -355,6 +355,12 @@ Blaze._materializeView = function (view, parentView, _workStack, _intoArray) {
       view.renderCount++;
       view._isInRender = true;
 
+      /**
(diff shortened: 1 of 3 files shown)


jankapunkt:
@DanielDornhardt can you specify some example code or (even better) add some tests that target specifically this code? We can also setup a meeting and do this together if you want.

jankapunkt:
@StorytellerCZ which branch should we target with this one?

PR #409, opened 2023-03-30
Feature/async blaze

### LINKED ISSUE #364: Handle async code
Blaze currently auto-calls functions:

```js
Template.example.helpers({
  asyncTest () {
    return {
      foo: () => 'bar'
    }
  }
})
```

```html
<template name="example">
  {{#with asyncTest}}{{this.foo}}{{/with}}
</template>
```

will lead to `bar` as output.

However, changing `foo` to an async function like so

```js
Template.example.helpers({
  asyncTest () {
    return {
      foo: async () => 'bar'
    }
  }
})
```

we will get for the same Template a `toString` output of a pending Promise: `[object Promise]`

However, we should be able to handle Promises in 2022 in any context. This would imply a huge PR but I think this is a crucial candidate for Blaze 3 and the future of Blaze

Grubba27:
This is a draft PR for people see how is adding async to blaze going.
We have accomplished unwrapping promises in the view.
Below there is a print of the screen + in the PR(will not be in the final version, there is the app that I'm using)
Note that to run it:

```bash
cd async-test
meteor npm i
npm run setup
meteor
```

Screenshot 2023-03-30 at 14 35 53
_foo is a result of a promise_

- [x] Add `#letAwait` to Blaze DSL with simple usage(just unwrap);
    - [ ] Tests 
- [ ] Work on having pending and error  states for `#letAwait` 
- [ ] Work in the building block such as `{{await something}}` or ``{{await s.foo}}``
- [ ] Work on having pending and error states for `await` keyword, maybe a #await block for this

--- 
I've made PR for release 3.0, but I can change this later. Just let me know :)

Closes #364


radekmie:
I've made some progress here:
* Prepared a sizeable suite of tests in a single template (see screenshot below).
    * No `{{#eachAwait}}` yet, but it looks like it'd be fairly simple. However, I think I like the truly low-level approach of `{{#letAwait}}` + `{{#each}}` more.
* Cleaned up the non-async implementation, to focus on _adding_ and not _changing_ code.
* Implemented `__async_state__` available only in the `{{#letAwait}}` blocks. It consists of two fields:
    * `error`, which is the first rejection or `undefined` when none was found.
    * `pending`, which indicates whether any of the promises is still pending.

image

radekmie:
Actually, we could handle it on a per-variable basis too:
```mustache
{{#letAwait x=foo}}
  Resolved value (if any): {{x}}.
  {{#if @pending 'x'}}x is still pending{{else}}x has resolved or rejected{{/if}}
  {{#if @rejected 'x'}}x has rejected {{@rejected 'x'}}{{else}}x has resolved or is still pending{{/if}}
  {{#if @resolved 'x'}}x has resolved {{@resolved 'x'}}{{else}}x has rejected or is still pending{{/if}}
{{/letAwait}}
```

It can handle general states (i.e., multiple variables) too:
```mustache
{{#letAwait x=foo y=bar z=baz}}
  Resolved value (if any): {{x}}, {{y}}, {{z}}.
  {{#if @pending}}some binding is still pending{{else}}all bindings have resolved or rejected{{/if}}
  {{#if @rejected}}some binding has rejected{{else}}all bindings have resolved or are still pending{{/if}}
  {{#if @resolved}}some binding has resolved{{else}}all bindings have rejected or are still pending{{/if}}
{{/letAwait}}
```

Of course these `@`-helpers can be switched to something like `Template.`, `this.`, or another. The idea is that the `#letAwait` binds the variables lazily and allows you to query their state. That'd mean we could simply make `#let` do that, as it'd behave the same for synchronous values (with some performance penalty, of course).

packages/spacebars-compiler/codegen.js
```diff
@@ -134,12 +135,12 @@ Object.assign(CodeGen.prototype, {
...
               self.codeGenInclusionData(args.slice(2)) +
               ', _variable: ' + BlazeTools.toJSLiteral(variable) + ' }; }';
-          } else if (path[0] === 'let') {
+          } else if (path[0] === 'let' || path[0] === 'letAwait') {
             var dataProps = {};
             args.forEach(function (arg) {
               if (arg.length !== 3) {
                 // not a keyword arg (x=y)
-                throw new Error("Incorrect form of #let");
+                throw new Error("Incorrect form of " + path[0]);
```
jankapunkt, line 143: Can we improve this error message early on? I'd rather have some more characters in the bundle but a clear error message that helps me in debugging when users report them.

jankapunkt: Okay my bad I just saw that all these views have these short error messages. We may discuss this separately.

radekmie: Yeah, all of them do that - I'd keep it as a separate topic.

radekmie:
As I'll be taking over this topic, I created #412 with an updated implementation.

@DanielDornhardt I'll address your comments later today or somewhere next week.

PR #412, opened 2023-05-05
Implemented async bindings in `#let`.

radekmie:
In this pull request, I made `#let` handle `Promise`s. It supersedes #409 as it solves the same problem, but there's no `#letAwait` needed.

### How it works

All Blaze `View`s store a binding mapping (`_scopeBindings`), where `#each` stores the `@index` variable and `#let` stores the locally scoped variables. Each binding is not a value but rather a `ReactiveVar` that is updated when needed.

In this PR, these `ReactiveVar`s no longer store the _value_ but rather a `Binding` object. Type-wise, it's either `undefined` (pending), `{ error }` (rejected), or `{ value }` (resolved). Synchronous values are immediately resolved (i.e., `{ value }` is used). The other states are reserved for asynchronous bindings (i.e., values wrapped with `Promise`s).

That means the following template:
```html
<template name="example">
  {{#let name=getName}}
    Hi, {{name}}!
  {{/let}}
</template>
```

Works with both synchronous and asynchronous `getName`:

```ts
Template.example.helpers({
  // Synchronous value.
  getName: 'John',

  // Asynchronous value.
  getName: Promise.resolve('John'),

  // Synchronous helper.
  getName: () => 'John',

  // Asynchronous helper.
  getName: async () => 'John',
});
```

### Async state

As the unwrapping of `Promise`s is not synchronous, all asynchronous values and helpers will start in a pending state. In such cases, the resolved value is always `undefined`. That means the template above will show `Hi, !` at first for both asynchronous examples. Similarly, a rejection will make the value `undefined`.

But there are cases where we'd like to know whether the operation is still pending or if it failed. To make it possible, there are three new global helpers:
* `@pending`, which checks whether any of the given bindings is still pending.
* `@rejected`, which checks whether any of the given bindings has rejected.
* `@resolved`, which checks whether any of the given bindings has resolved.

The usage looks as follows:

```html
<template name="example">
  {{#let name=getName}}
    {{#if @pending 'name'}}
      We are fetching your name...
    {{/if}}
    {{#if @rejected 'name'}}
      Sorry, an error occured!
    {{/if}}
    {{#if @resolved 'name'}}
      Hi, {{name}}!
    {{/if}}
  {{/let}}
</template>
```

All of them accept a list of names to check. Passing no arguments is the same as passing all bindings from the inner-most `#let`:

```html
<template name="example">
  {{#let name=getName}}
    {{#let greeting=getGreeting}}
      {{#if @pending}}
        We are fetching your greeting...
[...]

PR #413, opened 2023-05-05
Added support for `Promise`s in `Spacebars.call` and `Spacebars.dot`.

radekmie:
While #412 adds a way to unwrap `Promise`s in templates, here we'll focus on their usability. Currently, accessing a `Promise` results in `undefined` as they have very few accessible properties (most importantly, the resolved value is not there).

In this pull request, I made `Spacebars.call` and `Spacebars.dot` helpers aware of `Promise`s. The former resolves all of the arguments before calling the function, while the latter wraps property accesses in `Promise`s.

For details, please refer to the [newly added tests](https://github.com/meteor/blaze/pull/413/files#diff-88f4b0b1fc27941fdadab16c9d294823f271a74be57291b8dbde2750bb23d3aa).

_(Fun fact: there are 0 deleted lines in this pull request.)_

---

Once this and #412 will be merged, the "`await` unrolling" proposed before would not be needed. Consider the example from https://github.com/meteor/blaze/pull/409#issuecomment-1534702374:

```html
{{#if await someFunc ((await calcParamA) (await someOtherparamB)) (await calcParamB (await someOtherparamC) (someOtherparamD)) }}
    (...)
{{/if}}
```

This will have to unwrap the `Promise`, but the entire computation will be simple:

```html
{{#let result=(someFunc (calcParamA someOtherparamB) (calcParamB someOtherparamC someOtherparamD))}}
  {{#if result}}
    (...)
  {{/if}}
{{/let}}
```

Of course, it'll be possible to use the `@pending`, `@rejected`, and `@resolved` helpers to check the progress of calculating `result` as well.

### DIFF
--- a/packages/blaze/view.js
+++ b/packages/blaze/view.js
@@ -531,6 +531,12 @@ Blaze._isContentEqual = function (a, b) {
  */
 Blaze.currentView = null;
 
+/**
+ * @template T
+ * @param {Blaze.View} view
+ * @param {() => T} func
+ * @returns {T}
+ */
 Blaze._withCurrentView = function (view, func) {
   var oldView = Blaze.currentView;
   try {
--- a/packages/spacebars/spacebars-runtime.js
+++ b/packages/spacebars/spacebars-runtime.js
@@ -117,8 +117,48 @@ Spacebars.makeRaw = function (value) {
     return HTML.Raw(value);
 };
 
+// FIXME: Remove once `spacebars` can depend on `tracker@1.3.0` (Meteor 2.10.0).
+const _withComputation = Tracker.withComputation || function (computation, f) {
+  var previousComputation = Tracker.currentComputation;
+
+  Tracker.currentComputation = computation;
+  Tracker.active = !!computation;
+
+  try {
+    return f();
+  } finally {
+    Tracker.currentComputation = previousComputation;
+    Tracker.active = !!previousComputation;
+  }
+};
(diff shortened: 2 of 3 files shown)


DanielDornhardt:
```
{{#if await something.or.other }}
    (...)
{{/if}}
```

aaahhhh... noice... 

Couldn't we then not just implement `await` or something similar as a regular old global template helper then? 😀 

Something like this: 

```
Template.registerHelper('await', (value) => {
    const ourValue = new ReactiveVar()
    if (value instanceof Promise) {
        value.then((error, result) => {
            if(result) {
                ourValue.set(result)
            }
        })
    } 
    return ourValue.get()
})
```

... there's always the problem of the re-running of the autorun above then re-triggering the entire call which then would lead to re-triggering the call with a fresh promise which would then lead to... not the result which we want 😃 

-> But as a general idea it might be feasible and I'll toy with it when I'll have a bit of time. Or maybe @radekmie you know exactly how something like this could / should be implemented & do it in a heartbeat, which i'd be 100% onboard with :) 

Thank you for the PR & hear you later!

radekmie:
I wouldn't think of `await` at all for now, as it's unclear what should happen for the pending/rejected states - `#let` makes it obvious. I started a discussion whether `#if` and other blocks should handle it on Slack - we can discuss it there.

Grubba27 (approved):

PR #418, opened 2023-05-25
Fix reactivity for non-primitives

### LINKED ISSUE #417: Non-primitives not fully reactive in 2.7
Iterating nested arrays of objects from a reactive var don't get updated:

```html
<template name="myTemplate">
    {{#each entry in list}}
        {{! prints always: foo: 0}}
        {{entry.title}}: {{entry.users.length}}
        <button class="add-btn" data-index="{{@index}}">add users</button>
    {{/each}}
</template>
```

```js
import { Random } from 'meteor/random'

Template.myTemplate.onCreated(function () {
  const instance = this
  instance.nestedList = new ReactiveVar([{ title: 'foo', users: [] }])
})

Template.myTemplate.helpers({
  list () {
    const list = Template.instance().nestedList.get()
    console.debug(list) // will print correctly here!!!
    return list
  }
})

Template.myTemplate.events({
  'click .add-btn' (e, t) {
    const list = t.nestedList.get()
    list[0].users.push(Random.id())
    t.nestedList.set(list)
  }
})
```

Reverting to 2.6.x shows that it's introduced in 2.7

jankapunkt:
This fixes #417 to make nested data reactive again

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,4 +1,9 @@
-## v2.7.0, 2023-May-XX
+## v2.7.1, 2023-May-26
+
+* [#413](https://github.com/meteor/blaze/pull/418) Fix reactivity for non-primitives.
+
+
+## v2.7.0, 2023-May-23
 
 * [#413](https://github.com/meteor/blaze/pull/413) Added support for Promises in Spacebars.call and Spacebars.dot.
 * [#412](https://github.com/meteor/blaze/pull/412) Implemented async bindings in #let.
--- a/packages/blaze/.versions
+++ b/packages/blaze/.versions
@@ -3,7 +3,7 @@ babel-compiler@7.10.4
 babel-runtime@1.5.1
 base64@1.0.12
 binary-heap@1.0.11
-blaze@2.7.0
+blaze@2.7.1
 blaze-tools@1.1.3
 boilerplate-generator@1.7.1
 caching-compiler@1.2.2
@@ -28,7 +28,7 @@ htmljs@1.1.1
 id-map@1.1.1
 inter-process-messaging@0.1.1
 jquery@1.11.10
-local-test:blaze@2.7.0
+local-test:blaze@2.7.1
 logging@1.3.2
 meteor@1.11.2
 minimongo@1.9.3
--- a/packages/blaze/builtins.js
+++ b/packages/blaze/builtins.js
@@ -39,7 +39,12 @@ Blaze.With = function (data, contentFunc) {
  * @param {Binding} y
  */
 function _isEqualBinding(x, y) {
-  return x && y ? x.error === y.error && x.value === y.value : x === y;
+  if (typeof x === 'object' && typeof y === 'object') {
+    return x.error === y.error && ReactiveVar._isEqual(x.value, y.value);
+  }
+  else {
+    return ReactiveVar._isEqual(x, y);
+  }
 }
 
 /**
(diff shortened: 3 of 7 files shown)


StorytellerCZ (approved):

radekmie (commented):
So there are a couple of things:
1. The example you posted on Slack is, in my opinion, incorrect. I never thought that `ReactiveVar#set` **will trigger** when I set the exact same non-primitive value in there. I see it's documented, but that's _at least_ unexpected.
2. This code can be simplified in two steps:
    * Replace `undefined` with `{ value: undefined }` in `new ReactiveVar` in `Blaze._attachBindingsToView`.
    * Change `x && y ? x.error === y.error && x.value === y.value : x === y` to `ReactiveVar._isEqual(x.error, y.error) && ReactiveVar._isEqual(x.value, y.value)`.
3. We should add a test for that (maybe in a subsequent PR).
4. We should copy the example from Slack to here, as otherwise it'll disappear in 3 months.

jankapunkt:
> The example you posted on Slack is, in my opinion, incorrect. I never thought that ReactiveVar#set will trigger when I set the exact same non-primitive value in there. I see it's documented, but that's at least unexpected.

Incorrect in terms of that nested reactivity doesn't really exist with ReactiveVar? Then yes, it should be renamed to "fix reactivity for non-primitives"

> This code can be simplified in two steps:

I can add this to the PR

> We should add a test for that (maybe in a subsequent PR).

I can also add a test to this PR, I think it should not be that much additional effort

> We should copy the example from Slack to here, as otherwise it'll disappear in 3 months.

The example is already in the referenced issue #418

radekmie:
> > The example you posted on Slack is, in my opinion, incorrect. I never thought that ReactiveVar#set will trigger when I set the exact same non-primitive value in there. I see it's documented, but that's at least unexpected.
>
> Incorrect in terms of that nested reactivity doesn't really exist with ReactiveVar? Then yes, it should be renamed to "fix reactivity for non-primitives"

Well, yes, changing the name is fine. My thought was that I find it unintuitive that there's more logic than `===` 🤷 

> > This code can be simplified in two steps:
>
> I can add this to the PR

Great, thank you!

> > We should add a test for that (maybe in a subsequent PR).
>
> I can also add a test to this PR, I think it should not be that much additional effort

That would be awesome.

> > We should copy the example from Slack to here, as otherwise it'll disappear in 3 months.
>
> The example is already in the referenced issue #418 

Yeah, my bad, I checked the notifications in the reversed order.

radekmie (approved):

Grubba27:
tests need to be updated 🤔

jankapunkt:
Yes, I think the latest commit broke things. I'll check later and update.

jankapunkt:
The tests started failing after I changed 

```js
view._scopeBindings[name] = new ReactiveVar({ value: undefined }, _isEqualBinding);
```

it seems that at some other places the cope binding is not expecting this structure, mainly in all async tests. I can check if we can fix this, however, I would propose to revert to the previous commit, which worked stable.

radekmie (approved):
> it seems that at some other places the scope binding is not expecting this structure, mainly in all async tests. I can check if we can fix this, however, I would propose to revert to the previous commit, which worked stable.

I think it's something different, but this can be handled later. Let's ship this fix!

Grubba27:
If everyone is on board! will launch this patch

jankapunkt:
@Grubba27 go for it :+1:

Grubba27:
Blaze & spacebars are out 🚀

PR #419, opened 2023-06-02
fix: Failing tests

rodrigok:


jankapunkt:
Thanks a lot @rodrigok for this PR. We are always happy about new contributions. However, let me kindly ask you to respond to the following questions:

1. What is the goal of this pull request? I know it seems obvious but there should at least a short summary which tests are fixed in which situations.
2. You would also have to update the `test-app` to Meteor 3.0.0 in order to get the version conflicts resolved. Can you please do that?
3. Why `insertAsync`? from what I understand this is just a temporary API that will be removed once Meteor is fully migrated to async, @Grubba27 @StorytellerCZ is this assumption correct?

I'm putting this on `in-duscussion` until this is resolved. Please don't take this as an offense, but rather as a quality measure.

rodrigok:
@jankapunkt sure:

1. The goal is to fix the all the tests (but markdown ones, that I could not find a way to fix) for the Meteor's 3.0 branch where they are working to fix the tests https://github.com/meteor/meteor/pull/12658
2. Which versions? I'm pointing to the meteor:3 branch, I didn't change anything but failing tests, so I something related to versions should be changed it should be changed at the meteor-3.0 branch IMO
3. Since the objective is to fix the tests I don't believe it's relevant if it's temporary or not, a later PR may replace the APIs back keeping the awaits

jankapunkt:
@rodrigok 

2. is about the respective Meteor version, not the branch. We use a Meteor app under the `test.app` folder (see the [development guide](https://github.com/meteor/blaze/blob/master/CONTRIBUTING.md#run-local-tests))  that spawns the tests. If you want to fix the tests towards Meteor 3 then this app should have the Meteor 3 release in it. This is a separate repo so there is no Meteor 3 functionality unless explicitly configured.


> Since the objective is to fix the tests I don't believe it's relevant if it's temporary or not, a later PR may replace the APIs back keeping the awaits

I am not sure if I understood right but I thought that with Meteor's 3.0 release the temporary API will already be removed, which is why I thought we should skip the temp API at all.

denihs (approved):

denihs:
I'll go ahead and merge this so we can continue with the fixes.

PR #423, opened 2023-06-06
Replace spacebars.dot with rewritten version, fixing promises returning functions

DanielDornhardt:
Hi, the original "asyncified" `Spacebars.dot` didn't support functions returned from async functions... which might be a bit over the top, but supported is supported.

I think this should work as a drop-in replacement for the old `spacebars.dot`, but please review & check it. If this is not A++ feel free to take this as an encouragement to support functions returned from `async` steps during the `dot` resolution... :)

### DIFF
--- a/packages/spacebars/package.js
+++ b/packages/spacebars/package.js
@@ -1,7 +1,7 @@
 Package.describe({
   name: 'spacebars',
   summary: "Handlebars-like template language for Meteor",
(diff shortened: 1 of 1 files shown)


jankapunkt:
@DanielDornhardt what's the status here? Do you want to meet and discuss how we move on with it?

PR #424, opened 2023-06-21
Added support for `Promise`s in `#if` and `#each`.

radekmie:
In this pull request, I follow up on #412 and #413 by supporting `Promise`s in `#if`/`#unless` and `#each`. The idea is basically the same as how it works in `#let` except for where we actually store the `ReactiveVar`s.

To do:
- [x] **`#if`/`#unless` ~and `#each`~ are not rendering anything before the `Promise` resolves.** That's to make it clear that there's nothing there. I'm open for suggestions though, as I can imagine cases where `{{else}}` is more expected.
    * I asked a couple of people and they said it's fine not to render anything as in "not to make any decisions before the value is known".
    * Due to the complexity, `#each` will treat unresolved `Promise`s as empty sequences and thus render the `{{else}}` block. We may change that in the future.
- [x] **`#if`/`#unless` ~and `#each`~ are not rendering anything for rejected `Promise`s.** This is a sane choice, but I want to confirm that with the community.
    * I asked a couple of people and they said it's fine not to render anything as in "not to make any decisions before the value is known".
    * Due to the complexity, `#each` will treat rejected `Promise`s as empty sequences and thus render the `{{else}}` block. We may change that in the future.
- [x] Tests.
- [x] Documentation.
- [x] Decide whether or not we want to preserve the `if.__conditionVar` and `each.argVar`. They're not used anywhere, but _sometimes_ it makes sense (e.g., `#with` depends on that).
    * I decided to leave them for now, but there's quite a few potential simplifications in this area. They may improve the performance as well.

**Note:** #428 got merged into this branch!

### DIFF
--- a/packages/blaze/builtins.js
+++ b/packages/blaze/builtins.js
@@ -49,33 +49,50 @@ function _isEqualBinding(x, y) {
 }
 
 /**
- * Attaches a single binding to the instantiated view.
- * @param {Blaze.View} view Target view.
- * @param {string} name Binding name.
- * @param {unknown} value Bound value.
+ * @template T
+ * @param {T} x
+ * @returns {T}
  */
-function _setBindingValue(reactiveVar, value) {
+function _identity(x) {
+  return x;
+}
+
+/**
+ * Attaches a single binding to the instantiated view.
+ * @template T, U
+ * @param {ReactiveVar<U>} reactiveVar Target view.
+ * @param {Promise<T> | T} value Bound value.
+ * @param {(value: T) => U} [mapper] Maps the computed value before store.
+ */
+function _setBindingValue(reactiveVar, value, mapper = _identity) {
   if (value && typeof value.then === 'function') {
     value.then(
-      value => reactiveVar.set({ value }),
+      value => reactiveVar.set({ value: mapper(value) }),
       error => reactiveVar.set({ error }),
     );
   } else {
-    reactiveVar.set({ value });
+    reactiveVar.set({ value: mapper(value) });
   }
 }
 
 /**
+ * @template T, U
  * @param {Blaze.View} view Target view.
- * @param {unknown} binding Binding value or its getter.
+ * @param {Promise<T> | T | (() => Promise<T> | T)} binding Binding value or its getter.
  * @param {string} [displayName] Autorun's display name.
+ * @param {(value: T) => U} [mapper] Maps the computed value before store.
+ * @returns {ReactiveVar<U>}
  */
-function _createBinding(view, binding, displayName) {
+function _createBinding(view, binding, displayName, mapper) {
   const reactiveVar = new ReactiveVar(undefined, _isEqualBinding);
   if (typeof binding === 'function') {
-    view.autorun(() => _setBindingValue(reactiveVar, binding()), view.parentView, displayName);
+    view.autorun(
+      () => _setBindingValue(reactiveVar, binding(), mapper),
+      view.parentView,
+      displayName,
+    );
   } else {
-    _setBindingValue(reactiveVar, binding);
+    _setBindingValue(reactiveVar, binding, mapper);
   }
 
   return reactiveVar;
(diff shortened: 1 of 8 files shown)


DanielDornhardt:
Hi, i have some issues with our autoform forms not showing with this patch, previous version from Meteor 2.12 worked. I'll try to investigate what is happening.

Below is a screenshot of an error message I get, albeit only after a reload / HCP:

image

jankapunkt:
@DanielDornhardt this is likely an Autoform issue. If so please open in Autoform and we continue to investigate there

StorytellerCZ:
Would this be for Blaze 3 or 2?

DanielDornhardt:
Hi, i've made a small replication & also a loom video for a quick impression. I'll also cross-post in the `autoform` repository in case someone wants to help to identify & fix this issue.

radekmie:
Yeah, I see that it's not really `unless` now 😅 I'll push a fix for that soon and will take care of the rest this or next week.

DanielDornhardt:
Thank you @radekmie I'll check it out! :)

DanielDornhardt:
Lookin' better! :D Thank you a lot! We'll see how far this will get us!

sebastianspiller:
Hey @radekmie ,

I tried "if" in our project (I am a collegue of Daniel) and added it inside an html attribute like 

`<a class="{{#if isTruthyWithPromise}}arrow{{/if}}">`

This is not working and the browser is running on very high load.

radekmie:
@sebastianspiller I looked into it and unfortunately it's not that simple. The reason is the way Blaze handle attributes. Let's make an example:
```html
<template name="spacebars_async_tests_attributes">
  <img id="{{#if x}}1{{else}}0{{/if}}">
</template>
```

Translates to:
```ts
Template.__checkName("spacebars_async_tests_attributes");
Template["spacebars_async_tests_attributes"] = new Template("Template.spacebars_async_tests_attributes", (function() {
  var view = this;
  return HTML.IMG({
    id: function() {
      return Blaze.If(function() {
        return Spacebars.call(view.lookup("x"));
      }, function() {
        return "1";
      }, function() {
        return "0";
      })}
  });
}));
```

The culprit is that `id` is a function that creates a new `Blaze.If` every time - after its creation, it gets rendered once and destroyed (see `Blaze._expandView`). It's terribly inefficient but most importantly - stateless. And `Blaze.If` needs to be stateful now. So every time `Blaze.If` resolves its condition, it gets updated, destroyed and recreated - hence the loop and high CPU usage.

I tried to work around it, but there's no easy way around it. I'd even argue there's no way to fix it without touching the compiler and changing the way attributes are handled. And that's a lot.

DanielDornhardt:
Hi...

I mean - the `if`'s in attributes thing - I think that warrants some thinking... two things pop into my mind:

a) if that's an infinite loop, that probably should be prevented / the user should be warned if that's not supported
b) it also raises the question of the documentation & the consistency when async data sources are used within blaze templates - it'd be good to have a guide later to say "this stuff works, here are some workarounds we figured out, and these things aren't supported" in order for people to get a clear understanding of how it works. Eg here: https://www.blazejs.org/api/spacebars#In-Attribute-Values it doesn't warn users (yet) that this will crash the browser, we should update the docs after we're done.

c) maybe a + b can make an efficient developer rack his brain a bit whether a small compiler hack couldn't be warranted in order to avoid the busywork of both a) and b) 😄

radekmie:
I agree that for now, a detection + clear error would be needed for it to work. In the future, it could be improved by _somehow_ marking the attribute as async and for example wrap it with `#let`.

radekmie:
I just pushed some `Promise` detection and a couple of tests for that, so it won't silently loop forever. I also updated the docs.

@jankapunkt, @StorytellerCZ, @Grubba27 I consider this one done.

StorytellerCZ:
LGTM, what do you think @jankapunkt ?

StorytellerCZ (approved):

Grubba27:
Could not we target the devel(v2.x) branch for this PR?  This does not look to me as a breaking change.

radekmie:
Sure, I'll rebase it.

PR #432, opened 2023-11-27
Update dependencies in blaze-hot

StorytellerCZ:
Update dependencies and replace two `var` with `let`.
Also add some missing semi-colons.

### DIFF
--- a/packages/blaze-hot/hot.js
+++ b/packages/blaze-hot/hot.js
@@ -11,7 +11,7 @@ function patchTemplate(Template) {
   Template.registerHelper = function (name, func) {
     func[SourceModule] = currentModule.id;
     oldRegisterHelper(name, func);
-  }
+  };
 
   const oldOnCreated = Template.prototype.onCreated;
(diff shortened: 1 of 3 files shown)


packages/blaze-hot/hot.js
```diff
@@ -38,31 +38,31 @@ function patchTemplate(Template) {
...
 
     return oldOnDestroyed.call(this, cb);
-  }
+  };
 
   const oldHelpers = Template.prototype.helpers;
   Template.prototype.helpers = function (dict) {
     if (typeof dict === 'object') {
-      for (var k in dict) {
+      for (let k in dict) {
```
jankapunkt, line 46: please add a TODO note, that we should change this in the future to a solution that won't use for..in

jankapunkt (approved):

packages/blaze-hot/hot.js
StorytellerCZ: Probably can do this in this PR, I have fixed that in other PRs.

PR #433, opened 2023-11-27
Update caching-html-compiler

StorytellerCZ:
* Add missing import
* Add few missing `;`
* Add .jshintrc for code help with ES6 transition.
* Update dependencies

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,0 +1,4 @@
+{
+  "undef": true,
+  "esversion": 6
+}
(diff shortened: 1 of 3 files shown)


.jshintrc
```diff
@@ -0,0 +1,4 @@
+{
+  "undef": true,
```
jankapunkt, line 2: Do we want to drop jshint altogether?

jankapunkt (approved):

PR #436, opened 2023-11-27
Blaze-tool to ES6

StorytellerCZ:
Update dependencies of `blaze-tool`
Migrate old `var`s to `let` and `const`.
Simplified version (so far) of #397 

#397 had also some additional changes which I'm not that certain about.

### DIFF
--- a/packages/blaze-tools/tojs.js
+++ b/packages/blaze-tools/tojs.js
@@ -90,9 +90,9 @@ ToJSVisitor.def({
     }
 
     let attrsArray = null;
+    let needsHTMLAttrs = false;
     if (attrs) {
       attrsArray = [];
-      let needsHTMLAttrs = false;
       if (HTML.isArray(attrs)) {
         attrsArray = [];
         for (let i = 0; i < attrs.length; i++) {


packages/blaze-tools/tojs.js
```diff
@@ -89,19 +89,19 @@ ToJSVisitor.def({
       tagSymbol = 'HTML.getTag(' + toJSLiteral(name) + ')';
     }
 
-    var attrsArray = null;
+    let attrsArray = null;
     if (attrs) {
       attrsArray = [];
-      var needsHTMLAttrs = false;
+      let needsHTMLAttrs = false;
```
jankapunkt, line 95: needsHTMLAttrs is not defined in the tests. I assume this is a scoping issue and previously `needsHTMLAttrs` was simply wrong scoped and used outside it's defined scope and since it was a `var` this was possible.

I think this whole block of loops need some more refactoring effort. Maybe we can split it into several methods to better handle scoping?

StorytellerCZ: Agreed, I was going just for the minimum reproduction to move things forward one step at a time.

PR #438, opened 2023-11-28
Docs Hexo update

StorytellerCZ:
Updated Hexo to v6.3.
Currently unable to upgrade `hexo-renderer-less` as updating it breaks the look.


harryadel:
Just learnt about this from the latest [podcast episode](https://www.youtube.com/watch?v=avhkIo4kM-E), glad to see things moving for Blaze. Thank you Jan twins @jankapunkt @StorytellerCZ

PR #451, opened 2023-12-29
Feature/blaze tool es6

Grubba27:
https://github.com/meteor/blaze/pull/436 to the correct branch


jankapunkt:
@Grubba27 do we still want to merge this into the `2.8` branch?

packages/blaze-tools/tojs.js
```diff
@@ -89,19 +89,19 @@ ToJSVisitor.def({
       tagSymbol = 'HTML.getTag(' + toJSLiteral(name) + ')';
```
jankapunkt, line 89: let's use  template literals: 

```js
`HTML.getTag(${toJSLiteral(name)})`
```

jankapunkt (changes requested):
I would prefer template literals instead of concatenated strings. First, they increase readability (which is debatable in this case) and it coerces the values alaways to Strings, while `+` coerces values to primitives, incl. side effects.

PR #458, opened 2024-01-09
Blaze._expandAttributes returns empty object, if null or …

### LINKED ISSUE #456: TypeError: `Cannot convert undefined or null to object` in `waitForAllAttributesAndContinue`
### Summary
👋 After updating Blaze from **2.7.1** to **2.8.0** in our Meteor (v**2.14**) app, we got `Cannot convert undefined or null to object` error in `waitForAllAttributesAndContinue` function ([ref](https://github.com/meteor/blaze/blob/adc11c9a8bd5d638100d8ff7fb6dc3dc095671e4/packages/blaze/materializer.js#L100)):

image 28

In our case it happens when an HTML element has only one attribute and it's value is calculated by a Template helper function that returns `undefined`.

### Steps to reproduce
1. Create a Blaze app with command: `meteor create --blaze simple-todos-blaze --prototype`
2. Add `disabled` helper to `hello` Template in `main.js` module:
```js
Template.hello.helpers({
  disabled() {
    return undefined;
  },
```
3. Add `disabled` attribute to `button` element in `hello` Template in `main.html` module:
```js
<template name="hello">
  <button {{disabled}}>Click Me</button>
```
4. Run the app

#### Actual behaviour
The app crashes in browser with the described error

#### Expected behaviour
The element should be rendered with no attributes

jankapunkt:
- Fixes #456 
- tests added

### DIFF
--- a/packages/blaze/view.js
+++ b/packages/blaze/view.js
@@ -490,8 +490,9 @@ Blaze._expand = function (htmljs, parentView) {
 
 Blaze._expandAttributes = function (attrs, parentView) {
   parentView = parentView || currentViewIfRendering();
-  return (new Blaze._HTMLJSExpander(
+  const expanded = (new Blaze._HTMLJSExpander(
     {parentView: parentView})).visitAttributes(attrs);
+  return expanded || {};
 };
 
 Blaze._destroyView = function (view, _skipNodes) {
--- a/packages/spacebars-tests/template_tests.html
+++ b/packages/spacebars-tests/template_tests.html
@@ -430,6 +430,14 @@ Hi there!
   <p {{attrs}}></p>
 </template>
 
+<template name="spacebars_template_test_attr_object_helpers_are_disabled">
+    <button {{disabled}}>test</button>
+</template>
+
+<template name="spacebars_template_test_attr_object_helpers_are_disabled2">
+    <button title="foo" {{disabled}}>test</button>
+</template>
+
 <template name="spacebars_template_test_inclusion_helpers_are_isolated">
   {{> foo}}
 </template>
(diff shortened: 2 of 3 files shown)


radekmie (approved):

jankapunkt:
@StorytellerCZ @Grubba27 we should merge this asap, since there are people with breaking builds due to #456

radekmie:
Actually, I just checked and https://github.com/meteor/blaze/pull/460/files#diff-bf43372770fd37ca51fcb7ff881dee72730e4bb0a61e2a041c7ff61836c13aeaR107-R110 solves this issue as well as the async version of it (i.e., when the dynamic attribute return `Promise.resolve(undefined)` or similar). I suggest to merge #460 first and add only tests in this PR.

What do you think @jankapunkt?

radekmie:
Ping @jankapunkt.

jankapunkt:
@radekmie yes, let's merge #460 and once it's merged I will update this accordingly.

packages/blaze/view.js
```diff
@@ -490,8 +490,9 @@ Blaze._expand = function (htmljs, parentView) {
 
 Blaze._expandAttributes = function (attrs, parentView) {
   parentView = parentView || currentViewIfRendering();
-  return (new Blaze._HTMLJSExpander(
+  const expanded = (new Blaze._HTMLJSExpander(
     {parentView: parentView})).visitAttributes(attrs);
+  return expanded || {};
 };
```
jankapunkt, line 496: @radekmie this is the part that can be reset, because it's covered by your feature, right?

radekmie: Please double check it, but yes - it's covered in #460.

PR #474, opened 2025-01-06
Replace jquery parseHTML with native alternative

harryadel:
Decoupling jQuery and Blaze would take substantial effort. jQuery is used in many places:

1. HTML Parsing 
```js
DOMBackend.parseHTML = function (html) {
  return $jq.parseHTML(html, DOMBackend.getContext()) || [];
};
```
2. DOM Selection
```js
findBySelector: function (selector, context) {
  return $jq(selector, context);
}
```
3. Basic DOM Manipulation
```js
$jq(elem).remove();
$jq(elem).append(child);
```

4. Element Teardown Detection
```js
$jq.event.special[DOMBackend.Teardown._JQUERY_EVENT_NAME]
```

5. Event Delegation 
```js
delegateEvents: function (elem, type, selector, handler) {
  $jq(elem).on(type, selector, handler);
}
```

I've ranked them in terms of ease of replacement, and impact. The challenge lies mostly in testing and ensuring cross browser compatibility so it's best to merge each change individually, do a minor release, test then repeat until all is merged then we can do a major release.

I chose to start off with [parseHTML](https://api.jquery.com/jQuery.parseHTML/). It can already be replaced with [createHTMLDocument]( https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementation/createHTMLDocument) as suggested [here](https://youmightnotneedjquery.com/#parse_html). So I chose a middle approach where `createHTMLDocument` is used only if it succeeds with a fallback code that works with IE. 

I've relied on this [implementation](https://gist.github.com/Munawwar/6e6362dbdf77c7865a99) and the official [tests](https://github.com/jquery/jquery/blob/4466770992d5833358169d0248c4deedadea1a96/test/unit/core.js#L1352) in constructing the new tests to ensure backwards compatibility.

### DIFF
--- a/packages/blaze/dombackend.js
+++ b/packages/blaze/dombackend.js
@@ -30,22 +30,43 @@ DOMBackend.getContext = function() {
 }
 
 DOMBackend.parseHTML = function(html, context) {
-  if (!html) {
+  // Handle all falsy values and non-strings
+  if (!html || typeof html !== 'string') {
     return [];
   }
   
   context = context || DOMBackend.getContext();
-  html = html.trim();
   
-  // Return empty array for empty strings after trim
-  if (!html) {
+  // Return empty array for empty strings
+  if (!html.trim()) {
     return [];
   }
   
-  // Check if the string contains any HTML
-  if (!/(<|&(?:[a-z\d]+|#\d+|#x[a-f\d]+);)/i.test(html)) {
-    // Plain text, create a text node
+  // Check if the content contains any HTML
+  const hasHTML = /(<|&(?:[a-z\d]+|#\d+|#x[a-f\d]+);)/i.test(html);
+  
+  if (!hasHTML) {
+    // For pure text content, return a single text node
     return [context.createTextNode(html)];
+  }
+  
+  // Check for self-closing tag with content after
+  const selfClosingMatch = html.match(/^(<[^>]+\/>)([\s\S]*)$/);
+  if (selfClosingMatch) {
+    const [, tag, afterContent] = selfClosingMatch;
+    const result = [];
+    
+    // Parse the self-closing tag
+    const div = context.createElement('div');
+    div.innerHTML = tag;
+    result.push(div.firstChild);
+    
+    // Add content after as text node if present
+    if (afterContent) {
+      result.push(context.createTextNode(afterContent));
+    }
+    
+    return result;
   }
   
   // Handle special cases like <tr>, <td>, etc.
@@ -71,6 +92,11 @@ DOMBackend.parseHTML = function(html, context) {
   const firstTag = firstTagMatch ? firstTagMatch[1].toLowerCase() : null;
   const spec = firstTag ? specialParents[firstTag] : null;
   
+  // Split leading whitespace and content
+  const leadingMatch = html.match(/^(\s*)([^]*)$/);
+  const [, leadingWS, remainingContent] = leadingMatch;
+  
+  let contentNodes;
   try {
     // Try modern approach first
     if (context.implementation && context.implementation.createHTMLDocument) {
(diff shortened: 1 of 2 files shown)


harryadel:
cc @jankapunkt @radekmie @leonardoventurini @nachocodoner

jankapunkt (commented):
Thank you so much @harryadel for initiating this one!
I have one thing to discuss regarding the es5 code.

If Blaze depends on `ecmascript` then you're fine using es6+ because it will be compiled down (and for browser.legacy also in compatibility mode). On top of that, the classic `var` remains unscoped, while `let` and `const` are block-scoped, which allows for a more granular scoping and thus less errors.

However, this is just a suggestion and not a demand from my end and I'd like to see what others are saying.

harryadel:
I thought we need to support IE, no? that's why I opted for using es5 code. If that's not a concern I could revert back to es6

jankapunkt:
My understanding is, that `web.browser.legacy` builds in a way that IE is supported. However I'm not 100% if that's the case.

StorytellerCZ:
Plus IE is already out of official support anyway so I think we should scratch out IE support.
Right now the biggest offender is Safari.

packages/blaze/dombackend.js
```diff
@@ -14,27 +14,138 @@ DOMBackend.getContext = function() {
...
+  if (!html || typeof html !== 'string') {
+    return [];
+  }
+  
+  context = context || DOMBackend.getContext();
+  
+  // Return empty array for empty strings
+  if (html === "") {
+    return [];
+  }
```
radekmie, line 45: This is covered by `!html` above.

Is this implementation made by you or based on some existing one? It's not trivial (at least to me, because I'm not sure what "fancy stuff" is jQuery doing and what "IE quirks" are we talking about) and I'm wondering if there's maybe a different small and maintained library we could use instead.

Another difference I can see is that jQuery removes scripts by default ([see `keepScripts`](https://api.jquery.com/jQuery.parseHTML/)), which is not implemented here (and it's missing in the tests).

harryadel: >This is covered by !html above.

Good catch!

>Is this implementation made by you or based on some existing one? It's not trivial 

It's a mix of everything really, as I've said this [example](https://gist.github.com/Munawwar/6e6362dbdf77c7865a99) was source of inspiration along with jQuery's implementation.

> I'm wondering if there's maybe a different small and maintained library we could use instead.
There're definitely lots of libraries like [htmlparser2](https://github.com/fb55/htmlparser2) but again they're not drag n' drop kind of replacement. They require fine tuning to be backwards compatible. I'd love to be proven wrong if someone out there knows of a 1:1 alternative.

 Another difference I can see is that jQuery removes scripts by default ([see keepScripts](https://api.jquery.com/jQuery.parseHTML/)), which is not implemented here (and it's missing in the tests).
 
 >Another difference I can see is that jQuery removes scripts by default ([see keepScripts](https://api.jquery.com/jQuery.parseHTML/)), which is not implemented here (and it's missing in the tests).
 
 You're right. That can be added.
 
 
All in all, this PR can be used as a spring board for further discussion to seek our best solution. There're native APIs out there that can integrated like [createHTMLDocument](https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementation/createHTMLDocument) and [DOMParser](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser). Or using NPM libraries along with other modifications.

harryadel: @radekmie I made new modifications, now the code is a lot simpler and we also use https://github.com/apostrophecms/sanitize-html to handle XSS problems now. Let me know what you think.

distalx:
Should we consider use of `document.createDocumentFragment()`? It might help improve performance by reducing repaints when appending multiple child nodes. [ref](https://developer.mozilla.org/en-US/docs/Web/API/Document/createDocumentFragment#usage_notes).

jankapunkt:
@distalx I think we can put this on the list for improvements as it definitely makes sense for larger lists etc. However for now I'd like to have a maximum in compliance to the existing code behavior in order to not break things unless really necessary.

PR #475, opened 2025-02-03
Feature: addd explicit helper name + viewname to errors, causes…

jankapunkt:
As a follow up on the closed PR #368 I am providing a small fix to improve error context when thrown within helpers.

There are three states the error message can obtain:

- Lokal `'Exception in ${template.viewName} ${helperName}`'
- Global `'Exception in global helper ${name}'`
- Fallback `'Exception in template helper:'`


zodern (approved):
This looks very good.

It's probably out of scope for this PR, but I'm wondering if it would also be useful for some people to attach the details to the error object (error.helper, error.template, error.origionalMessge, etc.).

jankapunkt:
@zodern I personally would find this very helpful and I would go for another PR to add such feature to the existing error handler, which might even be the better solution for improved error handling without breaking existing things

PR #481, opened 2025-11-25
feat(blaze): Add visual error indicator for development

wreiske:
- Add error badge in bottom-left corner showing error count
- Add modal with detailed error information and stack traces
- Implement graceful failure with inline placeholders for missing templates
- Add error deduplication to prevent duplicate entries
- Integrate with HMR to auto-clear errors when templates are fixed
- Add accessibility features (ARIA labels, keyboard navigation)
- Auto-disable in production mode
- Add TypeScript definitions for new public API

### Runtime errors (missing templates)
`Spacebars.include` now catches errors from `templateOrFunction()` (e.g. `No such template: broken`) instead of letting them crash the entire page render. The error is reported to the error indicator, and an inline placeholder is rendered in place of the missing template. Other templates continue to render normally.

### Compile-time errors (invalid template syntax)
Spacebars parse errors (e.g. using `{{#if}}` blocks inside HTML attributes) no longer crash the Meteor build with "Errors prevented startup". Instead, the broken template is compiled into a fallback that renders an inline error placeholder with full details (file, line, error message). The error is also reported to `Blaze._errorIndicator` at runtime. A `console.warn` is emitted during build so the error remains visible in the server terminal.

BREAKING CHANGE: Missing template errors now show placeholders instead of throwing, which may change error handling behavior in some edge cases.

## Before Change

Before this PR, blaze template issues would just result in the entire app not working.

Screenshot 2025-11-25 at 1 23 27 AM

## After Change

Now the app works, AND it will gracefully continue rendering the rest of the app! Also a nice little error window to know where you messed up.

Screenshot 2025-11-25 at 1 22 11 AM
Screenshot 2025-11-25 at 1 22 22 AM


italojs (changes requested):

jankapunkt:
@wreiske I changed the merge base from `master` to `release-3.1.0` so we can ship this with the upcoming release. May you take a look at this again? The CI timeout should be fixed by now.

jankapunkt:
If anyone else wants to test and review this PR you need to checkout via

```shell
git fetch origin pull/481/head:feature/visual-error-indicator
git checkout feature/visual-error-indicator
```

packages/templating-tools/package.js
```diff
@@ -1,7 +1,7 @@
 Package.describe({
   name: 'templating-tools',
   summary: "Tools to scan HTML and compile tags when building a templating package",
-  version: '2.0.0',
+  version: '2.0.1',
```
jankapunkt: can you please reverse the version numbers? we will bump them on the alpha, beta, rc etc. for 3.1.0

packages/blaze/package.js
```diff
@@ -1,7 +1,7 @@
 Package.describe({
   name: 'blaze',
   summary: "Meteor Reactive Templating library",
-  version: '3.0.2',
+  version: '3.1.0',
```
jankapunkt: can you please reverse the version numbers? we will bump them on the alpha, beta, rc etc. for 3.1.0

packages/caching-html-compiler/package.js
```diff
@@ -2,7 +2,7 @@
 Package.describe({
   name: 'caching-html-compiler',
   summary: 'Pluggable class for compiling HTML into templates',
-  version: '2.0.0',
+  version: '2.0.1',
```
jankapunkt: can you please reverse the version numbers? we will bump them on the alpha, beta, rc etc. for 3.1.0

packages/html-tools/package.js
```diff
@@ -1,7 +1,7 @@
 Package.describe({
   name: 'html-tools',
   summary: "Standards-compliant HTML tools",
-  version: '2.0.0',
+  version: '2.0.1',
```
jankapunkt: can you please reverse the version numbers? we will bump them on the alpha, beta, rc etc. for 3.1.0

packages/spacebars-compiler/package.js
```diff
@@ -1,15 +1,15 @@
...
+  version: '2.0.1',
   git: 'https://github.com/meteor/blaze.git'
 });
 
 Package.onUse(function (api) {
   api.use('ecmascript@0.16.9');
 
   api.use('htmljs@2.0.1');
-  api.use('html-tools@2.0.0');
+  api.use('html-tools@2.0.1');
```
jankapunkt: can you please reverse the version numbers? we will bump them on the alpha, beta, rc etc. for 3.1.0

packages/spacebars-compiler/package.js
```diff
@@ -1,15 +1,15 @@
 Package.describe({
   name: 'spacebars-compiler',
   summary: "Compiler for Spacebars template language",
-  version: '2.0.0',
+  version: '2.0.1',
```
jankapunkt: can you please reverse the version numbers? we will bump them on the alpha, beta, rc etc. for 3.1.0

jankapunkt (changes requested):

jankapunkt:
@wreiske if you are busy I can also revert the versions if you want

wreiske:
Done — reverted all version number bumps in 15c3b0c. The five `package.js` files (`blaze`, `caching-html-compiler`, `html-tools`, `spacebars-compiler`, `templating-tools`) are back to their original versions, including the `api.use()` dependency references.

jankapunkt:
@wreiske I tested this using one of our medium sized apps that heavily uses dynamically imported templates and things work as expected except when adding faulty code in HMR mode:

When adding `<button {{#if counter}}disabled{{/if}}>Click Me</button>` to an already rendered template, I get no visual error indicators, but only the (server) console message 

```
=> Client modified -- refreshingWarning: Compile error in imports/ui/layout/navSide/navSide.html:14: Reactive HTML attributes must either have a constant name or consist of a single {{helper}} providing a dictionary of names and values.  A template tag of type BLOCKOPEN is not allowed here.
...  </a>       <button {{#if counter}}disab...
                        ^
The app will still run, but the affected template will show an error placeholder.
```

If I instead add something like `{{#if foo moo}}<p>hi</p>{{/if}}` , where `foo` does not exist, there is neither a visual error indicator, nor a console message.

Note, that this happens in a dynamically imported module and HMR active. The other behavior works well, even with dynamic imported modules.

packages/blaze/errorIndicator.js
```diff
@@ -0,0 +1,720 @@
...
+ * - Catches and displays Blaze/template-related errors
+ * - Graceful failure with inline error placeholders
+ * - Error deduplication
+ * - HMR integration for automatic error clearing
+ * - Accessible (ARIA labels, keyboard navigation)
+ * - Production mode detection (disabled by default in production)
+ * 
+ * @fileoverview Client-side error indicator for Blaze templates
+ */
+
```
jankapunkt, line 18: So I went through all the code here and the only question I have left is:

If we hard-fail on production builds, then wouldn't this file be obsolete? Meaning - Could we exclude this code if `Meteor.isProduction` to save the few KB for client load optimization?

**EDIT** - I maybe was a bit unspecific - while the code contains already checks for production mode, I think he majority of this file code be excluded at all on production as the other places have guard clauses that check for function existence before calling them.

I know with rspack etc. not an issue but I think we should, for backward compat, use the good ol' code splitting here, right?

Apart from that I would merge this and improve on request, once 3.1.0 alpha is out.

PR #483, opened 2025-12-12
docs: improvement of architecture documentation

jankapunkt:
This is ongoing work on an extensive architecture overview, see #482

Issue #468, opened 2024-08-18
#each re-rendered with stale data context


josherich:
Tested in Meteor 3.0.1
MacOS 14.6.1 
expected behavior: data context in view should be the same as currentData while re-rendering
actual behavior: data context in view is stale while re-rendering

Here's the minimal reprod using Blaze test code:
In the test code, the final `buf` shows
```
[
    "getMsgs called: foo",
    "view:foo, data:foo",
    "---flush---",
    "getMsgs called: bar",
    "view:foo, data:bar", <------------- foo in view, but bar in currentData
    "view:bar, data:bar"
]
```
I'm not 100% sure if this is a bug or expected behavior, but it can cause confusion since it's a very common pattern.
```js
Blaze.Template.registerHelper('$tpl', function() {
    return Blaze.Template.instance();
});
const R = ReactiveVar('foo');
const buf = [];

const eachTest = Blaze.Template('eachTest', function() {
    var view = this;
    return Blaze._TemplateWith(function() {
        return {
            foo: Spacebars.call(Spacebars.dot(view.lookup("$tpl"), "foo", "get"))
        };
    }, function() {
        return Spacebars.include(view.lookupTemplate("eachTestChild"));
    });
});
eachTest.onCreated(function() {
    const tpl = this;
    Object.assign(tpl, {
        foo: R,
    });
});

const eachTestChild = Blaze.Template('eachTestChild', function() {
    var view = this;
    return Blaze.Each(function() {
        return Spacebars.call(Spacebars.dot(view.lookup("$tpl"), "getMsgs"));
    }, function() {
        return ["\n    ", HTML.DIV("\n      Msg ", Blaze.View("lookup:msg", function() {
            return Spacebars.mustache(view.lookup("msg"));
        }), "\n      ", Blaze.View("lookup:$tpl.log", function() {
            return Spacebars.mustache(Spacebars.dot(view.lookup("$tpl"), "log"), view.lookup("msg"));
        }), "\n    "), "\n  "];
    });
});
Blaze.Template['eachTestChild'] = eachTestChild;
eachTestChild.onCreated(function() {
    const tpl = this;
    Object.assign(tpl, {
        getMsgs() {
            const foo = Template.currentData().foo;
            buf.push('getMsgs called: ' + foo);
            if (foo === 'foo') {
                return [{_id: '1', msg: 'foo'}];
            } else {
                return [{_id: '2', msg: 'bar'}];
            }
        },
        log(foo) {
[...]

Issue #473, opened 2024-11-29
Not rendering helper if collection "hidden" beside Promise


DblK:
If you make a simple async helper in conjuction with a meteor collection, during the sync of the collection, the helper is not updated.

I expect the helper to be refreshed once the collection is synced.

- Blaze: `3.0.1`
- Meteor: `3.0.2`

Here is a small code that you explain better the issue.
If you move up the line of `find.fetch` before the async call it's working as expected.
```js
Template.Home.onCreated(() => {
  Tracker.autorun(async () => {
    const user = await Meteor.userAsync();
    Meteor.subscribe('projects', [user?.someProp]); // Help sync Project Meteor Collection
  })
})

Template.Home.helpers({
  // This is not working as expected
  async someHelper(): Promise<string> {

    const someAsyncCall = await myFunction();
    if (someAsyncCall) {
      const projects = Project.find({}).fetch();
      console.log(projects.length) // Always 0
      return 'yeah';
    }

    return ''
  }
  // This is working as expected
  async someHelperWorking(): Promise<string> {

    const someAsyncCall = await myFunction();
    const projects = Project.find({}).fetch();
    if (someAsyncCall) {
      console.log(projects.length) // Always 10 (The amount subscribed)
      return 'yeah';
    }

    return ''
  }
})
```

Tell me if you need more information.

radekmie:
Can you create a repo with a reproduction we could run locally? That'd help immensely!

Issue #470, opened 2024-09-09
Bad resolve with template async parameter


DblK:
Blaze: `3.0.0` (`2.6.1` worked fine but no promise at that time)
Meteor: `3.0.2`

Since my migration of my application to Meteor 3.0, I figure out that some variable of template seems not to be resolve the way it used to be.

Here is an example :
```hbs
<template name="Test>
  {{#if myProp}}
     Yeah!
  {{else}}
    Dooh
  {{/if}}
</template>
```

If I use within blaze :
```hbs
{{#Test myProp=asyncProp}}
```

with :
```js
Template.XXX.helpers({
   async asyncProp() {
     return await something(); // return true
   }
})
```

The template will always display `Dooh` instead of `Yeah`.
I expected once the resolve had been done, to pass the new value to the template so it can revaluate it and change the display accordingly.

Did I miss something there?

radekmie:
Could you please provide a reproduction? Ideally a GitHub repo we could clone and run to see it.

DblK:
I did a quick and dirty new project to demonstrate the issue.
Here is the [repo](https://github.com/DblK/meteor-blaze-inline-template).

ddaydd:
Hi, where is the template "hello" ?

DblK:
My bad, to quick to make a repo example.
Now it should be clearer about the issue.

radekmie:
So, I checked it out and the issue is rather tricky. The point is that this...

```html
<img class={{x}}>
```

...translates to...

```typescript
Template("Template.test", (function() {
  var view = this;
  return HTML.IMG({
    class: function() {
      return Spacebars.mustache(view.lookup("x"));
    }
  });
}));
```

...which is fine. But if we'd add some text around, like this...

```html
<img class="1 {{#if x}}2{{else}}3{{/if}} 4">
```

...then it looks like this...

```typescript
Template("Template.test", (function() {
  var view = this;
  return HTML.IMG({
    class: function() {
      return [ "1 ", Blaze.If(function() {
        return Spacebars.call(view.lookup("x"));
      }, function() {
        return "2";
      }, function() {
        return "3";
      }), " 4" ];
    }
  });
}));
```

...which is a problem. The thing is that a new instance of `Blaze.If` is created every time `class` is executed. That makes it impossible to preserve the `__conditionVar` which stores the result of the `Promise`.

I'll think about it and I'd appreciate any ideas!

PR #492, opened 2026-03-07
release 3.1.0

jankapunkt:
This is our ongoing development toward the upcoming 3.1.0 release.

## Merged PRs

- [x] #489 
- [x] #483 
- [x] #487
- [x] #488 
- [x] #493
- [x] #478 
- [x] #495 
- [x] #494 
- [x] #497
- [x] #501
- [x] #513

## Resolved Issues

- [x] implement double testing (w jquery + w/o jquery) → #497
- [x] #468 → #501

## Release

- [x] 3.1.0-alpha
- [x] 3.1.0-beta
- [ ] 3.1.0-rc
- [ ] 3.1.0

## Summary by CodeRabbit

* **New Features**
  * Added JavaScript URL security controls via new Blaze security APIs
  * Enabled URL protocol caching for improved performance
  * Optimized attribute updates by tracking and reducing redundant handler calls

* **Improvements**
  * Made jQuery optional—Blaze now operates without it when unavailable
  * Removed lodash library dependencies in favor of native JavaScript equivalents
  * Expanded architecture documentation with comprehensive overview diagrams and terminology

* **Tests**
  * Added performance benchmarks for attribute updates and URL protocol detection


jankapunkt:
Note, I marked #470 and #473 as optional, moving them to the next 3.2 milestone, if no PRs coming in for them, as we should move towards an alpha release

jankapunkt:
@italojs @nachocodoner can one of you create an alpha `3.1.0-alpha` release of this branches latest commit?

// cc @dupontbertrand @StorytellerCZ

nachocodoner:
@jankapunkt: Done. `blaze@3.1.0-alpha.0` available with all other related packages all published.

dupontbertrand:
happening

dupontbertrand:
Status report on 3.1.0-alpha.0, with measurements, so this PR carries a clear picture of what is left.

## Verified green

Full Tinytest suite from `7d151bd8` (the commit alpha.0 was published from), run on **Meteor 3.5.2** rather than the 3.4 our CI test-app is pinned to:

| Configuration | Passed | Failed |
|---|---|---|
| native DOM backend | 414 | 0 |
| jQuery 3.7.1 backend | 418 | 0 |

Real-world: a substantial private compliance app of mine, in development and not launched yet, on Meteor 3.5.2 with published `blaze@3.1.0-alpha.0` and no `jquery` package (so native backend) passes 1172 server + 769 browser tests, 0 failures, identical to its score on Blaze 3.0.x. Ten further projects of mine run alpha.0 on Meteor 3.5.x.

## Blocking for a beta: #501 is optional, #513 is not

#513 fixes a regression the alpha actually ships. Proof that it is real and that our current suite does not cover it — the alpha with **only** #513's two tests added, fix withheld:

| Configuration | Passed | Failed |
|---|---|---|
| alpha.0 + #513 tests only, native | 414 | **2** |
| alpha.0 + #513 complete, native | 416 | 0 |
| alpha.0 + #513 complete, jQuery 3.7.1 | 420 | 0 |

The two failures are `delegated event fires in inclusion wrapped by an element inside {{#if}} (#512)` and `delegated event fires on direct child of a wrapper element (#512)`. Impact on my side: two silently dead controls on an admin screen while its eleven render tests stayed green.

## Proposed checklist for 3.1.0

- [ ] **#513** reviewed and merged — only known regression in the alpha, fix scoped to `dombackend.js`
- [ ] **#501** reviewed (#468 stale `#each` data context) — mergeable since June 6, long-standing bug rather than a 3.1 regression, but the fix is ready
- [ ] **#507** — milestone call, 3.1.0 or 3.2. It gates meteor/meteor#14355; see below
- [ ] #481 → 3.2 (changes requested), #509 → 3.2 (conflicting), #500 already 3.2
- [ ] `3.1.0-beta.0` version bump commit (I can prepare it)
- [ ] beta published, forum thread opened to collect app reports
- [ ] `3.1.0` published, `packages/non-core/blaze` submodule bumped in meteor/meteor

## #507 is a cross-repo decision, not a Blaze-internal one

#507 (server-side `Blaze.toHTML()`, the SSG/SSR foundation) is MERGEABLE against `release-3.1.0`, +278/-163 across 5 packages, zero reviews. It carries milestone 3.2 only because the 3.1 vs 3.2 call was left open in April and never resolved.

Meanwhile meteor/meteor#14355 (the `static-render` core package) is MERGEABLE and milestoned Release 3.6. @nachocodoner asked there, on July 29, for E2E scenarios on an existing E2E app with an SSR route. That E2E can only assert graceful degradation while `packages/non-core/blaze` is pinned to 3.0.3, because the server-side `Template` registry does not exist in Blaze 3.0.x.

So the two PRs are deadlocked on each other: #14355 waits on a Blaze release, #507 waits on a milestone decision.

Taking #507 into 3.1.0 closes the chain in the order the core team is already planning: Blaze 3.1.0 final → submodule bump in meteor/meteor → #14355 lands with real E2E coverage in 3.6. Leaving it for 3.2 parks #14355 behind a release with no date.

The cost, stated plainly: #507 is not a no-op. Lifting the `'client'` scoping puts `blaze` and `spacebars` in the server bundle and compiles `.html` for every architecture, for any app using `templating`, at roughly 500 bytes per template server-side. Backward compatible, documented in the PR's guide changes, but a real change of shape for a release currently framed around jQuery becoming optional. It also means the beta must be validated with #507 included, which is one more full run of the matrix above. I will do that run if the call goes that way.

## Two CI gaps worth fixing here

1. **Our test-app is pinned to `METEOR@3.4`.** The alpha is being used on 3.5.x in the wild and nothing covered that combination until now. Bumping it is the single highest-value CI change.
2. **When you bump it, the `test-app` job stops being native.** Since Meteor 3.5, `test-in-console` bundles `jquery` (meteor/meteor `70570e76f1`), and Blaze's weak `jquery` dependency loads it, so the job that is supposed to exercise the native backend silently runs jQuery back-compat instead. I had to use a local jquery-free copy of `test-in-console` to get a genuine native run. Worth asserting `Blaze._DOMBackend._hasJQuery === false` in that job so the configuration can never drift silently again.

Related, for meteor/meteor rather than here: all three Blaze E2E apps on `release-3.6` (`apps/blaze`, `apps/blaze-router`, `apps/full-blaze`) and all four Blaze-bearing skeletons pin `jquery`, so a core-side validation of 3.1 would also only exercise the jQuery path. One jquery-free variant would cover the native backend.

nachocodoner:
Thanks for the detailed report. I followed up with [meteor/meteor#14754](https://github.com/meteor/meteor/pull/14754), pointing the Blaze submodule to `7d151bd8` and adding your two delegated-event cases to an existing E2E app. They run with both native and jQuery backends, in development and production.

Locally, I can confirm the regressions: **8 tests pass and 4 fail**, with both cases failing only on the native backend in both modes. Applying only the `dombackend.js` change from #513 makes all 12 pass, and the full Blaze E2E group passes **36/36**. These are local results; CI is still pending. The PR keeps the original alpha pin so the failures remain visible. But we should definitely merge #513 for the next round and ensure [meteor/meteor#14754](https://github.com/meteor/meteor/pull/14754) gets passed.

Blaze submodule changes now trigger E2E, and all three Blaze apps verify and log the checked-out commit, resolved version, compiled source, and active backend. I think having this regression coverage in Meteor core is important for the stability of Blaze 3.1.0: it gives us repeatable checks through real apps and Meteor’s build pipeline, and keeps those cases covered as both projects evolve.

Also, I would add resolving this PR's conflicts between `release-3.1.0` and `master` to the proposed checklist.

Once the fixes are merged, the conflicts are resolved, and the updated submodule passes CI, I think that should be our signal to publish the first `3.1.0-beta.0` and communicate it to the community for broader testing.

nachocodoner:
Also, on your point about #507 being a decision across both repos, I agree we should plan the two changes together. I'm more in favor of keeping #507 for Blaze 3.2 and targeting **Meteor 3.7 or later** for [meteor/meteor#14355](https://github.com/meteor/meteor/pull/14355). That would also mean moving the Meteor PR out of the 3.6 milestone.

The [changes in #507](https://github.com/meteor/blaze/pull/507/files) remove the compiler's `archMatching: 'web'` restriction and make the `Template` runtime, `blaze`, and `spacebars` available on the server. As the guide in that PR explains, explicit imports control which templates reach the server, while eager-loaded templates add registration work at startup. Those are changes to existing apps using `templating`, even when they don't enable `static-render`. The PR estimates around 500 bytes per compiled template; we haven't measured the startup or memory impact ourselves.

The E2E results from [meteor/meteor#14754](https://github.com/meteor/meteor/pull/14754) give us a useful baseline for the current Blaze integration and the event regressions. They don't validate #507 or the new SSG/SSR path. I think that integration deserves its own E2E scenarios with real server-rendered routes, checks for existing apps, and measurements of the additional build and startup work.

For now, I think we already have enough to deliver with a stable Blaze 3.1.0 and the native backend. I'd focus the first beta on the regression fixes, resolving the branch conflicts, and getting broader app feedback. We can continue the SSG/SSR work toward the following release with time to validate both PRs together and keep the risk manageable.

jankapunkt:
@nachocodoner how do we move forward with test-in-console having jQuery added again?

nachocodoner:
> @nachocodoner how do we move forward with test-in-console having jQuery added again?

We will need to be a PR on the core again, and based on https://github.com/meteor/meteor/pull/14754 where we actually see the new Blaze 3.1.0 adopted as the default coming from the version. Hopefully we can deliver the whole thing for Meteor 3.6.

jankapunkt:
Sorry for being out of the loop lately, I had some private scrambles... What is the ETA for 3.6? Do we need a meeting to align ourselves? I am already reviewing a few things, I don't want to be the blocking factor here :sweat_smile:

dupontbertrand:
Thanks @nachocodoner, agreed on all points.

- **#507 / meteor/meteor#14355**: fine with Blaze 3.2 and Meteor 3.7+. I can't edit milestones on meteor/meteor, could you move #14355 out of *Release 3.6*?
- **#501**: updated with `release-3.1.0` now that #513 is in (merge commit). The only conflict was both PRs appending tests to the end of `spacebars-tests/template_tests.{html,js}`; both blocks kept. Green in native (420/0) and jQuery (424/0) on the pinned test-app release. @jankapunkt, ready for review whenever you have a moment.
- **meteor/meteor#14754 CI**: none of its 3 red jobs was related to #512. I opened meteor/meteor#14757 against that branch with the two fixes.
- **`release-3.1.0` ↔ `master` conflicts**: 34 files, no code. They come from the 3.0.3 release: version bumps in `package.js` / `.versions` (all lower than the 3.1.0-alpha versions), `HISTORY.md`, and `site/package-lock.json`. Happy to open a PR merging `master` into `release-3.1.0` (keeping the 3.1.0 versions and carrying the 3.0.3 history entry) if that helps.

jankapunkt:
> release-3.1.0 ↔ master conflicts: 34 files, no code. They come from the 3.0.3 release: version bumps in package.js / .versions (all lower than the 3.1.0-alpha versions), HISTORY.md, and site/package-lock.json. Happy to open a PR merging master into release-3.1.0 (keeping the 3.1.0 versions and carrying the 3.0.3 history entry) if that helps.

Yes that would help a lot :+1:

jankapunkt:
#515 is merged, so is #511 however the merge conflict seems to remain :thinking:

dupontbertrand:
My bad, I should have flagged this on #515: it needed a merge commit. The squash kept the resolved content but dropped `master` as a second parent, so git still sees the 3.0.3 commits as unmerged and #492 reports the same conflicts.

The content on `release-3.1.0` is already right, so the fix only records the history, with no file change. `-s ours` is safe here because #515 already brought every `master` change over. Two options:

- **Directly**, since you have push rights:
  ```
  git checkout release-3.1.0 && git pull
  git merge -s ours origin/master -m "Merge master into release-3.1.0 (record history only)"
  git push origin release-3.1.0
  ```
- **Or via #516**, which contains exactly that commit (0 files changed). Please merge it with **Create a merge commit**, not squash.

I checked locally: the resulting tree is identical to `release-3.1.0`, `master` becomes an ancestor, and #492 no longer conflicts.

jankapunkt:
@dupontbertrand @nachocodoner  I moved everything but #501 to 3.2 so this is the remaining last item to review and merge toward 3.1.0

nachocodoner:
Could we do a documentation-gap review before merging this release?

A few examples I noticed:

- The migration guide says jQuery was removed, but the implementation makes it optional and continues using it as the DOM backend when installed.
- The Templates API says `TemplateInstance.$` always returns a jQuery object, while without jQuery it returns an array of DOM elements.
- The JSDoc and TypeScript declarations do not accurately represent both return types.
- Some migrated pages still contain Hexo-specific directives, such as {% apibox %}, {% dtdd %}, and the changelog tag, which VitePress does not process correctly.
- The documentation build and deployment should be verified in CI to catch these issues.

These should be reviewed so the documentation accurately explains the new optional-jQuery behavior and the migrated site renders correctly.

Also, lets keep coderabbitai comments resolved or closed, but not remaining open.

jankapunkt:
@nachocodoner I would address these remaining points, do you think it makes sense to publish a beta in parallel? I think most of these do not involve significant code changes so I think a beta + announcement in the forums for early adoption tests can make sense, what do you think?

nachocodoner:
Yes, that is indeed the next step for it. We will do it next.

PR #497, opened 2026-03-31
Tests: jquery + native DOM backends

### LINKED ISSUE #353: Remove jQuery
This has been mentioned from time to time and I'd like to bring this up to seriously discuss it.

Removing jQuery brings some big advantages

- people tend to avoid Blaze, simply because it's the only dependency to jQuery
- reduce bundle size by around 70-90 kb
- do not rely to fix security issues that come with jQuery
- use browser-builtin standards do not introduce new dependencies or complex custom implementations

There are also some caveits:

- the DOMBackend basically relies on[ jquery to parse HTML](https://github.com/meteor/blaze/blob/master/packages/blaze/dombackend.js#L18) although this might seem to [not be that hard to replace](https://github.com/jquery/jquery/blob/main/src/core/parseHTML.js), this still needs a lot of testing and especially integration testing, since this is basically the backbone of Blaze IMO
- we should first merge https://github.com/meteor/blaze/pull/334
- we need to preserve API, because there are some builtins that heavily work similar to jQuery (like `templateInstance.$`) 

Also there is the question of versioning and if the jQuery removal would really have a breaking impact, which would force us towards a major bump.

What do you think?

jankapunkt:
This is a follow up on #493 which adjusts the tests to be compatible with and without jQuery.

- [x] remove all direct jQuery dependencies in Blaze packages
- [x] remove jquery dependencies in `test-in-browser` (attention, [this is a core package](https://github.com/meteor/meteor/blob/devel/packages/test-in-browser/))
- [x] pass all tests with jQuery
- [x] pass all tests without jQuery
- [x] fix Teardown for native DOM backend
- [x] fix event delegation for native DOM backend
- [x] adapt CI to run testsuite with both setups in sequence (first using native, then jQuery)

## Findings form initial research

📢 Update: all of the issues below were addressed in this PR, see comments further below

**1. tests satisfaction**
This PR focused only on passing tests when DOM backend is native AND if jQuery is used.

**2. jQuery in tests**
Many tests used jquery by default, for example to find or remove elements:

```js
$(div).remove()
```
This causes the native DOM backend to fail, because there is no native communication between the removal and the Teardown (which removes reactive listeners etc.).
As a solution, the native `HTMLElement.prototype.remove` method gets overridden to call the teardown for the element and its children.
**This needs to be tested appropriately using apps.**

**3. template scoped events**
With jQuery, events are automatically scoped to the template level. This prevents event selector in a child template to bubble up the event to the parent template, if the selector matches.
As a solution, the parent chain is traversed upwards to see if a view is associated and if it changed, once the selector does not match the current element anymore.
**This needs to be tested appropriately using apps.**

**4. external dependencies**
The weak dependency of jQuery automatically loads, once any other package has loaded jQuery. However, the `test-in-browser` (peer-dep of Tinytest via tinytest-harness) causes jQuery to be included in any case.
**This is therefore on hold, until `test-in-browser` has jQuery removed.**

Once this is done, we can leverage an env variable to include jQuery in tests, without adding it as a dependency and adjust CI to run tests with jQuery and with native DOM backend.


jankapunkt:
@dupontbertrand do you have some time to spare to take a look at this one?

dupontbertrand (commented):
Incoming..

packages/blaze/materializer.js
```diff
@@ -97,7 +97,7 @@ const isPromiseLike = x => !!x && typeof x.then === 'function';
 
 function then(maybePromise, fn) {
   if (isPromiseLike(maybePromise)) {
-    maybePromise.then(fn, Blaze._reportException);
+    maybePromise.then(fn).catch(Blaze._reportException);
```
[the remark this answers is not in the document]
dupontbertrand: I think it's better to revert to .then(fn, Blaze._reportException) (two-arg form). The current .then(fn).catch(Blaze._reportException) has two issues: (1) isPromiseLike accepts any thenable (typeof x.then === 'function'), but arbitrary thenables are not guaranteed to return an object with .catch() from their .then() — this can throw at runtime; (2) .catch() also swallows errors thrown inside fn, which is a behavior change from the original two-arg form.

packages/test-in-browser/package.js
```diff
@@ -0,0 +1,39 @@
...
+});
+
+Npm.depends({
+  'bootstrap': '5.3.8',
+  'diff': '8.0.2'
+});
+
+Package.onUse(function (api) {
+  api.use('ecmascript');
+  // XXX this should go away, and there should be a clean interface
```
dupontbertrand: I'd suggest removing the orphan XXX comment copied from meteor/meteor's original package. Since this is a fresh temporary stub, carrying over legacy TODOs adds noise

packages/spacebars-tests/template_tests.js
```diff
@@ -2587,14 +2589,19 @@ Tinytest.add(
     });
 
     tmpl.rendered = function () {
-      $(this.find('button')).trigger('someCustomEvent', args);
+      if (hasJquery) {
+        $(this.find('button')).trigger('someCustomEvent', args);
+      } else {
```
dupontbertrand: Dead code: the "jQuery.trigger extraParameters" test is already wrapped in if (hasJquery) at L2572, but the test body contains another if (hasJquery) { ... } else { trigger(...) }. The else branch is unreachable. I think the inner conditional should be removed, keeping only the jQuery path

packages/spacebars-tests/old_templates_tests.js
```diff
@@ -162,13 +166,17 @@ Tinytest.add(
...
+      : div.querySelectorAll(':scope > *');
     test.equal(elems.length, 0);
     test.equal(canonicalizeHtml(div.innerHTML), 'asdf');
 
     R.set('<span class="hi">blah</span>');
     Tracker.flush();
-    elems = $(div).find('> *');
+    elems = hasJquery
+      ? $(div).find('> *')
+      :  div.querySelectorAll(':scope > *');
```
dupontbertrand: Extra space before div.querySelectorAll

packages/spacebars-tests/old_templates_tests.js
```diff
@@ -2784,11 +2802,21 @@ Tinytest.add(
...
+      $(div.querySelector('.toremove')).remove();
+    } else {
+      div.querySelector('.toremove').remove();
+    }
     assertCallsAndListeners(0, 0, 0, 0);
   }
 );
 
+
+const trigger = (el, eventType, bubbles = true) => {
```
dupontbertrand: Inconsistent trigger helper signatures: positional args (el, eventType, bubbles) vs destructured ({ el, eventType, bubbles, options }). Both are new code. I think it's better to pick one form and use it in both files` template_tests.js (L3333)`

packages/spacebars-tests/template_tests.js
```diff
@@ -3309,11 +3321,21 @@ Tinytest.add(
...
+    if (hasJquery) {
+      $(div.querySelector('.toremove')).remove();
+    } else {
+      div.querySelector('.toremove').remove();
+    }
     assertCallsAndListeners(0, 0, 0, 0);
   }
 );
 
+const trigger = ({ el, eventType, bubbles = true, options }) => {
```
dupontbertrand, line 3329: Inconsistent trigger helper signatures: positional args (el, eventType, bubbles) vs destructured ({ el, eventType, bubbles, options }). Both are new code. I think it's better to pick one form and use it in both files` old_templates_tests.js (L2815)`

jankapunkt:
@dupontbertrand all comments are addressed in 6c208492 and 79a52996

packages/spacebars-tests/old_templates_tests.js
```diff
@@ -2833,11 +2862,21 @@ Tinytest.add(
           'You might need to defocus the Chrome Dev Tools to get a more accurate run of this test!',
       });
       borken = true;
-      $(input).trigger('focus');
+      if (hasJquery) {
+        $(input).trigger('focus');
+      } else {
+        trigger({ el: input, eventName: 'focusin', bubbles: true });
```
dupontbertrand: It's eventName or eventType here ?

packages/spacebars-tests/old_templates_tests.js
```diff
@@ -2833,11 +2862,21 @@ Tinytest.add(
...
+      }
     }
     test.equal(buf.join(), 'FOCUS');
     blurElement(div.querySelector('input'));
-    if (buf.length === 1) $(input).trigger('blur');
+    if (buf.length === 1) {
+      if (hasJquery) {
+        $(input).trigger('blur');
+      } else {
+        trigger({ el: input, eventName: 'focusout', bubbles: true });
```
dupontbertrand: It's eventName or eventType here ?

packages/spacebars-tests/old_templates_tests.js
```diff
@@ -2852,7 +2891,13 @@ Tinytest.add(
     Tracker.flush();
     test.equal(div.querySelectorAll('input').length, 1);
     focusElement((input = div.querySelector('input')));
-    if (borken) $(input).trigger('focus');
+    if (borken) {
+      if (hasJquery) {
+        $(input).trigger('focus');
+      } else {
+        trigger({ el: input, eventName: 'focusin', bubbles: true });
```
dupontbertrand: It's eventName or eventType here ?

jankapunkt:
@dupontbertrand thanks, fortunately this mistake occurred and you detected it, because the current tests never branched into any of the manual event triggers, because the `focusElement` / `blurElement` worked out fine.

I therefore added in 46edae1a some explicit tests that only cover manual event dispatching of focus/blur

dupontbertrand (approved):

PR #499, opened 2026-04-03
fix: improve error messages for invalid template tags in HTML start tags

dupontbertrand:
## Summary

- The pattern `<input {{#if isDisabled}}disabled{{/if}}>` causes a **silent compiler crash** with a misleading error: `Cannot find module '../.././imports/ui/.../template.html'` and `Template.xxx is not defined`. Developers waste hours debugging a module resolution issue that is actually a Spacebars syntax problem.
- This PR replaces the cryptic failure with specific, actionable error messages for each invalid pattern:
  - `{{#if}}`/`{{#unless}}`/`{{#each}}`/`{{#with}}` in start tag: suggests the `attribute="{{helper}}"` pattern with Bad/Good syntax example
  - `{{{triple-stache}}}` in start tag: explains the raw HTML issue
  - `{{> inclusion}}` in start tag: clear rejection message
  - Custom block helpers: specific message mentioning dictionary helpers
- Add test coverage with string assertions for each error path

### Before

```
Cannot find module '../.././imports/ui/pages/myPage/template.html'
Template.myTemplate is not defined
```

### After

```
{{#if}} is not allowed in an HTML start tag. To conditionally add a boolean
attribute like "disabled", "checked", or "selected", use the
attribute="{{helper}}" syntax instead.
  Bad syntax:  <input {{#if isDisabled}}disabled{{/if}}>
  Good syntax: <input disabled="{{isDisabled}}">
```

## Test plan

- [x] Tested manually with a Meteor 3.4 app using local package symlinks
- [x] Verified `{{#if}}` standalone attribute shows new message with Bad/Good example
- [x] Verified `{{{triple-stache}}}` in start tag shows raw HTML message
- [x] Verified `{{> inclusion}}` in start tag shows inclusion message
- [x] Verified valid case (`{{#if}}` inside attribute value) still compiles and runs
- [x] Tinytest suite (spacebars-compiler - parse) passes — 416/416 tests, 0 failures

### DIFF
--- a/packages/spacebars-compiler/spacebars_tests.js
+++ b/packages/spacebars-compiler/spacebars_tests.js
@@ -249,25 +249,25 @@ Tinytest.add("spacebars-compiler - parse", function (test) {
   // Triple-stache in start tag — should mention raw HTML
   test.throws(function () {
     SpacebarsCompiler.parse('<a {{{x}}}></a>');
-  }, /Triple-stache.*raw HTML/);
+  }, 'not allowed in an HTML start tag');
 
   // Block helper in start tag — should suggest attribute="{{helper}}" pattern
   test.throws(function () {
     SpacebarsCompiler.parse('<a {{#if x}}{{/if}}></a>');
-  }, /not allowed in an HTML start tag/);
+  }, 'not allowed in an HTML start tag');
 
   // Common real-world patterns that trigger this error
   test.throws(function () {
     SpacebarsCompiler.parse('<input {{#if isDisabled}}disabled{{/if}}>');
-  }, /Bad syntax.*Good syntax/);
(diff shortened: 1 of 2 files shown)


jankapunkt:
@dupontbertrand @wreiske can you please check if this may conflict with pr #481

wreiske:
I believe [#481](https://github.com/meteor/blaze/pull/481) does cover this use case, yes.

Screenshot 2026-04-08 at 9 02 36 AM

dupontbertrand:
My bad, don't know something like that was already there ! @wreiske @jankapunkt

wreiske:
No worries at all! It shows that multiple people thought it needed improvement so there's some validation there! ☺️

PR #500, opened 2026-04-03
feat: inline expressions in Spacebars

### LINKED ISSUE #355: Support logical operators in HTML Template code
I often have much redundancy when just trying to pass conditional paramters. For example:

````html
{{#if condition}}
  {{> otherTemplate param=1}}
{{else}}
  {{> otherTemplate param=2}}
{{/if}}
```

alternative would be now to create another helper (because `condition` is shared, which I can't simply return the value from this helpert) but that increases the js code a lot for many simple conditions.

Way better would be

````html
{{> otherTemplate param=(condition ? 1 : 2)}}
```

That would also make many custom Template helpers packages obsolete and thus again reduce bundle size.

Any ideas or ciritcs on that?

### LINKED ISSUE #459: Allowing arbitrary expressions
[Allow arbitrary expressions](https://edgejs.dev/docs/introduction) is a useful future as I find myself constantly creating helpers to do basic computations using [veliovgroup/Meteor-Template-helpers](https://github.com/veliovgroup/Meteor-Template-helpers).

dupontbertrand:
## Summary

- Add inline expression support to Spacebars, allowing arithmetic, comparison, logical, and ternary operators directly in template tags
- Eliminates the need for helper functions in common cases like `{{price + tax}}`, `{{#if score >= 90}}`, `{{isActive ? "on" : "off"}}`
- **100% backward compatible** — existing `{{helper arg}}` syntax is unchanged

### Supported operators

| Category | Operators |
|----------|-----------|
| Arithmetic | `+`, `-`, `*`, `/`, `%` |
| Comparison | `===`, `!==`, `>`, `>=`, `<`, `<=` |
| Logical | `&&`, `\|\|`, `!` |
| Ternary | `? :` |
| Grouping | `( )` |
| Unary | `-` (negation), `!` (logical not) |

### Examples

```html
{{price * quantity + tax}}
{{#if items.length > 0}}...{{/if}}
<div class="{{isActive ? 'on' : 'off'}}">
{{score >= 90 ? "A" : score >= 80 ? "B" : "C"}}
{{(a + b) * c}}
```

### How it works

The parser uses a **Pratt parser** (top-down operator precedence) that integrates with the existing scanner. After scanning the first path, it peeks for an operator — if found, it switches to expression mode. Since operators like `+`, `*`, `===`, `&&` never appear in valid Handlebars, there is no ambiguity with existing syntax.

- `templatetag.js` — Pratt parser with JS-standard operator precedence
- `codegen.js` — recursive `codeGenInlineExpr()` for code generation
- `spacebars-runtime.js` — lightweight `Spacebars.expr()` post-processor

### Known limitation

`{{!foo}}` is still parsed as a Spacebars comment. Use `{{(!foo)}}` as a workaround.

### Forbidden operators (with clear error messages)

`|` (reserved for future pipes), `&`, `++`, `--`, `=`, `+=`

## Test plan

- [x] Tested with a 14-section Blaze app covering all expression types
- [x] Arithmetic: `+`, `-`, `*`, `/`, `%`, chained, literals, negative multiplier
- [x] Comparison: all 6 operators, string/boolean literals, mixed with logical
- [x] Logical: `&&`, `||`, `!` prefix, negated groups
- [x] Ternary: simple, nested, with arithmetic in branches
- [x] Grouping: precedence, single value, double parens, multiple groups
- [x] Unary: `-a`, `a + -b`, `a - -b`, `(-a) * (-b)`
- [x] Dot access: `items.length > 0`, `items.length * 10`
- [x] Block helpers: `{{#if expr}}`, `{{#unless expr}}`, `{{#if a + b > 10}}`
- [x] Attribute values: ternary in class/style, arithmetic in data-attr, input value
- [x] Whitespace: `{{a+b}}`, `{{ a + b }}`, `{{a +b}}`
- [x] Reactivity: expressions update when reactive sources change
- [x] Dynamic templates: `Template.dynamic`, `#each item in items`, `Template.contentBlock`
- [x] **Backward compat**: simple paths, dot paths, helper calls, negative args, sub-expressions, keyword args, `{{this}}`, `{{@index}}`, `{{{raw}}}`, `#let`, global helpers

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+## vNEXT
+
+### Highlights
+* Inline expressions in Spacebars: use arithmetic (`{{a + b}}`), comparison (`{{a > b}}`), logical (`{{a && b}}`), and ternary (`{{a ? "yes" : "no"}}`) operators directly in templates without helper functions. Fully backward compatible.
+
 ## v3.0.2, 2025-02-04
 
 ### Highlights
--- a/packages/spacebars-compiler/codegen.js
+++ b/packages/spacebars-compiler/codegen.js
@@ -310,7 +310,12 @@ Object.assign(CodeGen.prototype, {
       break;
     case 'EXPR':
       // The format of EXPR is ['EXPR', { type: 'EXPR', path: [...], args: { ... } }]
-      argCode = this.codeGenMustache(argValue.path, argValue.args, 'dataMustache');
+      if (argValue.expr) {
+        // Inline expression inside sub-expression: {{helper (a + b)}}
+        argCode = this.codeGenInlineExpr(argValue.expr);
+      } else {
+        argCode = this.codeGenMustache(argValue.path, argValue.args, 'dataMustache');
+      }
       break;
     default:
       // can't get here
--- a/packages/spacebars-compiler/codegen.js
+++ b/packages/spacebars-compiler/codegen.js
@@ -67,6 +67,28 @@ const makeObjectLiteral = (obj) => {
 };
 
 Object.assign(CodeGen.prototype, {
+  // Generate JavaScript code for an inline expression AST node.
+  // Each PathExpression is resolved via view.lookup() and unwrapped
+  // via Spacebars.call() so that reactive helpers are evaluated.
+  codeGenInlineExpr: function (node) {
+    switch (node.type) {
+    case 'BinaryExpression':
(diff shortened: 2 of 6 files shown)


jankapunkt:
@dupontbertrand I would not consider this part of 3.1.0 to keep the release manageable and constraint the surface of what needs to be field tested (and fixed if people report issues).

If it's breaking and not 100% backwards compatible then we should target 4.0.0 otherwise it should target 3.2.0.

I will create branches for both releases this weekend.

PR #506, opened 2026-04-16
Blaze 3.0.3

nachocodoner:
This PR prepares the next Blaze patch, which includes:

* [[#480](https://github.com/meteor/blaze/pull/480)](https://github.com/meteor/blaze/pull/480) Fix SWC compatibility by transforming `arguments` into rest parameters (`...args`)

The idea is to release it with Meteor 3.4.1. We will open a testing period of about a week, and then proceed with it alongside the official 3.4.1 release.

I'm not very familiar with Blaze releases. I've only done one direct official release before. But looking at the history, it seems we basically bump all packages and the relationships between them. I'm not sure that is necessary for every kind of change. Normally, it would make sense to bump only the packages that changed, and maybe the related ones too. Still, I believe bumping everything is the safest approach, so that's what I'm doing. Let me know if there is anything else I should know.


jankapunkt:
@nachocodoner this is correct, the versions used to be bumped for all packages so users get the full update when updating `blaze-html-templates` which is the "meta-package" that initiates the dependency chain.

I think Denilson published them using the publish shell script provided in this repo. Let me know when `blaze-html-templates@3.0.1-rc.0` is available then I will do some tests with our apps for backwards compatibility with 3.4.0 and 3.0.2.

nachocodoner:
I published `blaze@3.0.3-rc.0` and all the related packages. The improvement to make it compatible with SWC seemed to work on my test.

image

nachocodoner:
All E2E tests for Blaze and the manual testing performed showed this change is working and is not risky to publish.

I've published Blaze 3.0.3 and all related packages so I can link the official release to the upcoming Meteor 3.4.1 official version.

I'm going to merge this PR already to master to link it properly with Meteor core.

PR #507, opened 2026-04-20
feat: enable server-side Blaze template rendering (foundation for SSG/SSR)

dupontbertrand:
## Summary

Enables `Blaze.toHTML()` to work on the server by making compiled templates and the `Template` registry available server-side. This is the foundation for SSG/SSR use cases (demonstrated in a working POC — see links below).

The key insight is that `Blaze.toHTML()` already works server-side through a DOM-free code path — the only problem was that compiled templates and the `Template` global were restricted to the client. This PR lifts that restriction without breaking anything client-side.

## Changes

5 small, backward-compatible changes across 5 packages:

| Package | Change |
|---------|--------|
| `templating-compiler` | Remove `archMatching: 'web'` so templates compile for all architectures |
| `templating-tools` | Guard `generateBodyJS()` with `Meteor.isClient` — body rendering uses `document.body` |
| `caching-html-compiler` | Guard body attrs setter with `Meteor.isClient` |
| `templating-runtime` | Export `Template` to server, guard DOM code with `Meteor.isClient` |
| `templating` | Export `Template` to server (matches `templating-runtime`) |

## Why this is safe

- **`<template>` tag compiled output** is pure JavaScript (`Template.__checkName` + `Template[name] = new Template(...)`). No DOM access.
- **`<body>` tag compiled output** uses `document.body` — wrapped in `Meteor.isClient` to stay client-only.
- **`Template.__checkName`, `Template._migrateTemplate`, `Blaze.Template` constructor** are already server-safe (no DOM dependencies).
- **DOM-related code** (`Template.body`, HMR, `_applyHmrChanges`) wrapped in `Meteor.isClient` to remain client-only.
- **Server-side overhead**: ~500 bytes per compiled template. For an app with 100 templates, ~50 KB of JS loaded at startup.

## What this does NOT change

- Client-side rendering behavior is unchanged
- `blaze`, `htmljs`, `spacebars` cores are unchanged
- The DOM-dependent code (`materializer`, `domrange`, `dombackend`, `events`, `attrs`) stays client-only

## Proof of concept

Built on top of these changes:
- Demo: https://ssg-ssr-demo.sandbox.galaxycloud.app (classic Meteor build)
- Demo: https://ssg-ssr-demo-rspack.sandbox.galaxycloud.app (rspack build, via a small patch to `@meteorjs/rspack` — separate PR)
- Source: https://github.com/dupontbertrand/meteor-ssg-ssr-demo

Forum discussion: https://forums.meteor.com/t/ssg-ssr-for-meteor-blaze-a-proof-of-concept/64556

## Testing done

- Existing Blaze usage (client-side rendering) continues to work unchanged
- `Template.myTemplate` is now accessible server-side after `import '../lib/templates.html'` in `server/main.js`
- `Blaze.toHTML(Template.myTemplate)` returns proper HTML strings server-side
- `Blaze.toHTMLWithData(Template.myTemplate, data)` works with data contexts
- `#each`, `#if`, `#unless` render correctly through `_expandView(forExpansion=true)`
- Production `meteor build` succeeds
- Vanilla Blaze apps (no SSG/SSR usage) build and run identically

## Consumer-side usage

Apps that want to use server-side rendering can now do:

```js
// server/main.js
import '../lib/templates.html';  // makes Template.* available server-side

// later — pure server-side rendering
const html = Blaze.toHTMLWithData(Template.myTemplate, { title: 'Hello' });
```

Templates must be written as pure render functions against explicit data — no `Session`, no `this.subscribe()`, no `Template.dynamic` (which remains client-only).

### DIFF
--- a/packages/caching-html-compiler/caching-html-compiler.js
+++ b/packages/caching-html-compiler/caching-html-compiler.js
@@ -115,14 +115,17 @@ CachingHtmlCompiler = class CachingHtmlCompiler extends CachingCompiler {
         }
       });
 
-      // Add JavaScript code to set attributes on body
+      // Add JavaScript code to set attributes on body.
+      // Guarded with Meteor.isClient because document.body doesn't exist on the server.
       allJavaScript +=
-`Meteor.startup(function() {
-  var attrs = ${JSON.stringify(compileResult.bodyAttrs)};
-  for (var prop in attrs) {
-    document.body.setAttribute(prop, attrs[prop]);
-  }
-});
+`if (Meteor.isClient) {
+  Meteor.startup(function() {
+    var attrs = ${JSON.stringify(compileResult.bodyAttrs)};
+    for (var prop in attrs) {
+      document.body.setAttribute(prop, attrs[prop]);
+    }
+  });
+}
 `;
     }
 
--- a/packages/templating-compiler/compile-templates.js
+++ b/packages/templating-compiler/compile-templates.js
@@ -1,7 +1,8 @@
 /* global CachingHtmlCompiler TemplatingTools */
 Plugin.registerCompiler({
   extensions: ['html'],
-  archMatching: 'web',
+  // archMatching removed — compile for all architectures (web + os/server)
+  // so that Template definitions are available server-side for SSG rendering.
   isTemplate: true,
 }, () => new CachingHtmlCompiler(
   'templating',
--- a/packages/templating-runtime/package.js
+++ b/packages/templating-runtime/package.js
@@ -11,9 +11,10 @@ Package.onUse(function (api) {
   // XXX would like to do the following only when the first html file
   // is encountered
 
-  api.export('Template', 'client');
+  // Export Template to both client and server — server needs it for SSG rendering.
+  api.export('Template');
 
-  api.addFiles('templating.js', 'client');
+  api.addFiles('templating.js');
 
   // html_scanner.js emits client code that calls Meteor.startup and
   // Blaze, so anybody using templating (eg apps) need to implicitly use
@@ -27,13 +28,14 @@ Package.onUse(function (api) {
     'meteor',
     'blaze@3.0.0',
     'spacebars@2.0.0'
-  ], 'client');
+  ]);
 
   // to be able to compile dynamic.html. this compiler is used
   // only inside this package and it should not be implied to not
   // conflict with other packages providing .html compilers.
   api.use('templating-compiler@2.0.0');
 
+  // dynamic template support is client-only (requires DOM)
   api.addFiles([
     'dynamic.html',
     'dynamic.js'
--- a/packages/templating-runtime/package.js
+++ b/packages/templating-runtime/package.js
@@ -11,9 +11,10 @@ Package.onUse(function (api) {
   // XXX would like to do the following only when the first html file
   // is encountered
 
-  api.export('Template', 'client');
+  // Export Template to both client and server — server needs it for SSG rendering.
+  api.export('Template');
 
-  api.addFiles('templating.js', 'client');
+  api.addFiles('templating.js');
 
   // html_scanner.js emits client code that calls Meteor.startup and
   // Blaze, so anybody using templating (eg apps) need to implicitly use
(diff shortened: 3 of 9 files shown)


jankapunkt:
I would target this for release 3.2.0 to keep 3.1.0 manageable and finish it soon

dupontbertrand:
As you want boss, to be honest it's a "small" changes but if you prefer to wait for 3.2 it's up to you, maybe you have to check with someone else because this PR have to be merged at the same time at this one  https://github.com/meteor/meteor/pull/14355

dupontbertrand:
Cross-reference so the discussion stays in one place: the sequencing question for this PR is being discussed in meteor/meteor#14355, where @nachocodoner asked for E2E coverage of the `static-render` package. That E2E can only assert graceful degradation until this lands and the `packages/non-core/blaze` submodule in meteor/meteor moves off 3.0.3 — so the 3.1.0 vs 3.2 milestone call decides it. @jankapunkt the question is waiting for you there.

I measured the rebase in the meantime, and it is small: `release-3.1.0` is 2 commits ahead of the merge base, and the only real conflict is `packages/templating-runtime/package.js`, where the version bumps (`blaze@3.0.0` → `3.1.0-alpha.0`) sit adjacent to the `], 'client')` line this PR removes. Resolution is "take both sides", about 6 lines. `packages/templating/package.js` auto-merges. I'll clear it regardless of the milestone call — it presumes nothing.

One thing to fix in the same pass: this PR adds `site/source/guide/server-rendering.md` without a matching entry in `site/.vitepress/config.mts`, so the page is unreachable from the guide sidebar.

dupontbertrand:
Pushed one more docs commit.

The guide claimed that "if you don't import `.html` files from your server entry point, nothing changes". That is false for two changes this PR makes unconditionally to every app using `templating`:

- `blaze` and `spacebars` join the **server bundle** — the `'client'` scoping on `templating-runtime`'s `api.use` was the only thing excluding them, and it has to be lifted for the registry to exist server-side.
- removing `archMatching: 'web'` compiles `.html` for the server too, so an app that **eager-loads** (no `imports/` directory) registers every template on the server at startup.

Neither renders anything by itself, and rendering stays opt-in behind an explicit `Blaze.toHTML()` call — but both cost server bundle size and startup work, and someone deciding whether to upgrade needs that stated. The section now distinguishes explicit-imports apps from eager-loading ones.

Still missing here, and I would rather name it than leave it implicit: **there are no tests for the server-availability feature itself**. The only test change is a scanner expectation update. Nothing asserts that `Template.foo` exists on the server or that `toHTMLWithData` renders on the `os` arch. I can add those if you want them in this PR rather than a follow-up.

jankapunkt (commented):
@dupontbertrand the code looks good to me. Would you mind adding a few SSG/SSR tests so we have this covered, too?

jankapunkt:
@dupontbertrand one more thing, I moved this to 3.1 due to discussion in #492

dupontbertrand:
Thanks @jankapunkt. Just to make sure we're aligned: in #492, Nacho proposed keeping #507 for 3.2 and moving meteor/meteor#14355 to Meteor 3.7+ ([comment](https://github.com/meteor/blaze/pull/492#issuecomment-5663280461)), and I agreed. His reasoning: #507 makes `blaze`, `spacebars` and the `Template` runtime available on the server for every app using `templating`, so he'd like dedicated E2E scenarios and build/startup measurements first, while beta.0 focuses on the native-backend fixes.

Did you move it to 3.1 on purpose, meaning you'd rather ship it in 3.1.0? If so, I'll rerun the full matrix with #507 included before the beta. Otherwise it can go back to 3.2.

jankapunkt:
woops sorry I misunderstood that. so I moved it back to 3.2

dupontbertrand:
@jankapunkt added in dff8e0d5: server rendering tests in `templating-runtime`.

- a compiled template renders to the expected HTML
- data is escaped in text and attributes (`<script>` in `{{title}}`, `"` and `&` in an attribute), while `{{{raw}}}` is left as-is
- block helpers: `#if`/`#else`, `#unless`, `#each`, `#each … in`, `#with`
- helpers and inclusion arguments
- a server-loaded `.html` file that also contains a `<body>` with attributes

The first four run on both client and server with the same hand-written expected strings, so they also check that both sides render identically.

How I checked that they actually cover this PR:

| Run | Meteor 3.4 (test-app pin) | Meteor 3.5.2 |
|---|---|---|
| New tests on `release-3.1.0`, without this PR | 5 server tests fail (`Blaze is not defined`), 4 client tests pass | – |
| Full suite with this PR, native | 425 passed, 0 failed | 425 passed, 0 failed |
| Full suite with this PR, jQuery | 429 passed, 0 failed | 429 passed, 0 failed |
| `Meteor.isClient` guard removed in `generateBodyJS` | server crashes at load (`reading 'addContent'`) | – |
| Body-attributes guard removed in `caching-html-compiler` | server crashes at boot (`document is not defined`) | – |

On 3.5.2, the native run uses a local copy of `test-in-console` without `jquery`, since the 3.5 driver pulls jquery in and would otherwise turn it into a second jQuery run.

Not covered on purpose: async helpers (there is no `toHTMLAsync`) and `Template.dynamic`, which stays client-only.

